PR #9, opened 2018-07-16
Higher-res and pixel grid-aligned versions of the ChRIS plugin store logo

mairin:
i aligned all of the elements of the chris store logo to the pixel grid so they should appear crisp and clear now. also updated with a vector version of the chris brain logo.
resolves https://github.com/FNNDSC/cube-design/issues/12


mairin:
screenshot from 2018-07-16 14-35-41

The resolution can't be increased any more without some changes to the HTML and I'm not sure we want to do those (would require making the black bar up top taller, which I think is probably not a good idea.)

jdtzmn:
@jcaianirh I made some changes to the HTML and CSS to support the higher-res logos. After edits/approval I will squash the commits and merge.

PR #15, opened 2018-07-23
Added placeholders for loading content

jdtzmn:
Previously, pages that were loading looked like this:

old placeholder

Now they look like this:

new placeholders


src/components/Plugins/Plugins.js
```diff
@@ -91,20 +94,36 @@ class Plugins extends Component {
...
-      );
+      pluginListBody = new Array(6).fill().map((e, i) => (
+        // eslint-disable-next-line react/no-array-index-key
+        <LoadingPluginItem key={i} />
+      ));
     }
 
-    const pluginListLength = pluginList ? pluginList.length : 0;
+    let pluginsFound;
+    if (pluginList) {
```
jcaianirh: repeating conditional check here. you can do it all in the same conditional above.

jdtzmn: Good catch! I'll update the files accordingly.

PR #16, opened 2018-07-23
Removed auth-token requests for plugins

jdtzmn:


src/components/Plugins/Plugins.js
```diff
@@ -44,19 +44,15 @@ class Plugins extends Component {
 
   fetchPlugins() {
     const storeURL = process.env.REACT_APP_STORE_URL;
-    const authURL = process.env.REACT_APP_STORE_AUTH_URL;
+    const client = new StoreClient(storeURL);
```
jcaianirh: remove storeURL const 
newStoreClient(process.env.REACT_APP_STORE_URL); 
its not being used anyplace else.

jdtzmn: @jcaianirh I thought I'd leave it for readability sake; it was intentional. I can remove it though if you want.

jcaianirh: no its ok to leave if you think its more readable.

PR #17, opened 2018-07-26
Sign In page

jdtzmn:
Screenshot

### With error:

Screenshot with error

### With patternfly error:

Screenshot with pattern fly error

### With patternfly layout

screen shot 2018-08-06 at 4 48 21 pm

### DIFF
--- a/src/components/SignIn/SignIn.js
+++ b/src/components/SignIn/SignIn.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { Redirect, Link } from 'react-router-dom';
 import {
-  Card, CardTitle, CardBody, Button, Icon,
+  Card, CardTitle, CardBody, Button, Alert,
   Form, FormGroup, FormControl, ControlLabel,
 } from 'patternfly-react';
 import { StoreClient } from '@fnndsc/chrisstoreapi';
@@ -68,10 +68,8 @@ class SignIn extends Component {
     this.setState({ error: message });
   }
 
-  hideError(e) {
-    if (!e.key || e.key === 'Enter') {
-      this.setState({ error: null });
-    }
+  hideError() {
+    this.setState({ error: null });
   }
 
   render() {
@@ -89,16 +87,13 @@ class SignIn extends Component {
           {
             error && (
               <div className="signin-error-container">
-                <div
+                <Alert
                   className="signin-error"
-                  role="button"
-                  tabIndex={0}
-                  onClick={this.hideError}
-                  onKeyUp={this.hideError}
+                  type="error"
+                  onDismiss={this.hideError}
                 >
-                  <Icon name="times-circle-o" />
-                  {` ${error}`}
-                </div>
+                  {error}
+                </Alert>
               </div>
             )
           }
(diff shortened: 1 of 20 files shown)


src/components/SignIn/SignIn.js
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  hideError(e) {
+    if (!e.key || e.key === 'Enter') {
+      this.setState({ error: null });
+    }
+  }
+
+  render() {
+    if (this.state.toDashboard) {
```
jcaianirh: const {toDashboard, error, username, password, loading} = this.state;

src/components/SignIn/SignIn.js
```diff
@@ -0,0 +1,166 @@
...
+
+    this.handleUsername = this.handleUsername.bind(this);
+    this.handlePassword = this.handlePassword.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.showError = this.showError.bind(this);
+    this.hideError = this.hideError.bind(this);
+  }
+
+  componentWillMount() {
+    this.mounted = true;
```
jcaianirh: Whats going on here? Do we need to keep track of the mounted state?

src/components/SignIn/SignIn.js
```diff
@@ -0,0 +1,166 @@
...
+    if (this.state.toDashboard) {
+      return <Redirect to="/developers" />;
+    }
+
+    return (
+      <div className="signin">
+        <div className="signin-container">
+          {
+            this.state.error && (
+              <div className="signin-error-container">
```
jcaianirh: do you mind showing an image of the error case in the pr description please?

src/components/SignIn/SignIn.js
jdtzmn: It has to do with memory leaks (very small ones of course) in React. The Promise will keep running (not easy to stop it), but there is no reason to re-render a nonexistent component when it redirects to the developers page. This was the simple fix I found.

src/components/SignIn/SignIn.js
jdtzmn: Done!

src/components/SignIn/SignIn.js
```diff
@@ -0,0 +1,166 @@
...
+
+  componentWillMount() {
+    this.mounted = true;
+  }
+
+  componentWillUnmount() {
+    this.mounted = false;
+  }
+
+  handleUsername(event) {
```
jcaianirh: onChange(event) {
  this.setState({ 
    [e.target.name]: e.target.value 
  });
//Add name to the form control 
}

src/components/SignIn/SignIn.js
jcaianirh: Thanks @jdtzmn! Would you mind using the pfly react alert to handle these errors?

https://rawgit.com/patternfly/patternfly-react/gh-pages/index.html?selectedKind=patternfly-react%2FWidgets%2FAlert&selectedStory=Alert%20types&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

jdtzmn: Done!

jcaianirh:
I'd really like this to follow the patternfly pattern for Basic Login located here if possible: 

https://www.patternfly.org/pattern-library/application-framework/login-page/#code

We don't need to add the parts we don't plan on using (Language, Keep me logged in.., etc.).

jdtzmn:
@jcaianirh Done!

PR #21, opened 2018-09-19
Add 404 Page not found for invalid page routes.

jcaianirh:
Add 404 page designed by Shania. Add route logic, image, tests and snapshot.

screen shot 2018-09-19 at 4 55 57 pm

### DIFF
--- a/src/components/NotFound/NotFound.test.js
+++ b/src/components/NotFound/NotFound.test.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import NotFound from './NotFound';
+
+describe('NotFound', () => {
+  let wrapper;
+  beforeEach(() => {
+    wrapper = shallow(<NotFound />);
+  });
+
+  it('should render correctly', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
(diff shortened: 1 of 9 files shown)


src/components/NotFound/NotFound.test.js
```diff
@@ -0,0 +1,21 @@
...
+  let wrapper;
+  beforeEach(() => {
+    wrapper = shallow(<NotFound />);
+  });
+
+  it('should render correctly', () => {
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render  div', () => {
```
jdtzmn: You might want to specify the div “not-found div” instead of just “div” or specify that you are checking the img

jdtzmn:
@jcaianirh thank you for continuing to write tests!

jcaianirh:
Thanks @jdtzmn for the review!! Will def change the test description. I think for the button, since the page is wrapped in the header and footer we should be ok with linking back to content. Good to hear from you!!!

mairin:
Hey I saw this assigned to me, do you need anything from me?

jcaianirh:
@mairin all set, Jacob reviewed the pr.

PR #29, opened 2018-11-19
Add developer dashboard to chris store ui.

jcaianirh:
Add a developer dashboard to the chris store ui based on designs in the design repo. Provide links to the existing Add Plugin Page.

screen shot 2018-11-18 at 8 54 18 pm


jcaianirh:
@jbernal0019 yes just the empty state is implemented so far.

Issue #36, opened 2018-11-29
are you sure?


jcaianirh:
provide are you sure? for delete action in dashboard page

gagansk:
Closed with https://github.com/FNNDSC/ChRIS_store_ui/pull/45

PR #45, opened 2019-02-15
Issue36 fixed using Patternfly-React Confirm Alter box

gagansk:
Issue https://github.com/FNNDSC/ChRIS_store_ui/issues/36 has been fixed.

Changes:
1. A dependency is added to support React Confirm Alter dialog in package.json
2. The dialog functionality is handled in the deletePlugin function. On clicking yes in the dialog box, the Plugin delete API will be called. On clocking no, an later box will be displayed to let the user know the plugin was no deleted.

@jcaianirh , Please review the Pull request.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,4 @@
 {
-  "parser": "babel-eslint",
   "extends": "airbnb",
   "env": {
     "node": true,
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,4 +1,5 @@
 {
+  "parser": "babel-eslint",
   "extends": "airbnb",
   "env": {
     "node": true,
--- a/src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
+++ b/src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
@@ -64,7 +64,7 @@ class DashPluginCardView extends Component {
   state = {
     show: false,
   };
-  deletePlugin = (name) => {
+  deletePlugin(name) {
     const { onDelete } = this.props;
     onDelete(name);
     this.setState(() => ({ show: false }));
@@ -81,8 +81,6 @@ class DashPluginCardView extends Component {
     let pluginCardBody;
     const { plugins } = this.props;
     const showEmptyState = isEmpty(plugins);
-    const primaryContent = <p className="lead">Are you sure?</p>;
-    const secondaryContent = <p>Plugin will be permanently deleted</p>;
     const addNewPlugin = (
       <Col xs={12} sm={6} md={4} key="addNewPlugin">
         <Card>
@@ -105,6 +103,8 @@ class DashPluginCardView extends Component {
       pluginCardBody = plugins.map((plugin) => {
         const creationDate = new RelativeDate(plugin.creation_date);
         const applicationType = new DashApplicationType(plugin.type);
+        const primaryContent = <p className="lead">Are you sure?</p>;
+        const secondaryContent = <p>Plugin <b>{plugin.name}</b> will be permanently deleted</p>;
         return (
           <Col xs={12} sm={6} md={4} key={plugin.name}>
             <Card>
--- a/src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
+++ b/src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
@@ -12,12 +12,10 @@ import {
   Card,
   CardHeading,
   DropdownKebab,
-  MenuItem,
   FieldLevelHelp,
+  MessageDialog,
 } from 'patternfly-react';
-import { confirmAlert } from 'react-confirm-alert';
 
-import 'react-confirm-alert/src/react-confirm-alert.css';
 import './DashPluginCardView.css';
 import BrainImg from '../../../../assets/img/empty-brain-xs.png';
 import PluginPointer from '../../../../assets/img/brainy_welcome-pointer.png';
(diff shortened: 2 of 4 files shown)


gagansk:
PR comment given by @jcaianirh  has been addressed.

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -61,10 +61,21 @@ class DashPluginCardView extends Component {
 
     this.deletePlugin = this.deletePlugin.bind(this);
   }
+  state = {
+    show: false,
```
jcaianirh: Lets be more descriptive about this. maybe showConfirmation

.eslintrc
```diff
@@ -1,4 +1,5 @@
 {
+  "parser": "babel-eslint",
```
jcaianirh: what is this? do we need this?

gagansk: Yes. Since we are using the statement, "state = ", eslint throws an error due to incompatibility between your development environment and ESLint's current parsing capabilities with the ongoing changes with JavaScripts ES6~7. This line solves the incompatibility issue. 

Ref:
[https://stackoverflow.com/questions/36001552/eslint-parsing-error-unexpected-token]

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -61,10 +61,21 @@ class DashPluginCardView extends Component {
...
+  };
   deletePlugin(name) {
     const { onDelete } = this.props;
     onDelete(name);
+    this.setState(() => ({ show: false }));
```
jcaianirh: this.setState({ show: false }); should be good here.

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -61,10 +61,21 @@ class DashPluginCardView extends Component {
...
   }
+  secondaryAction = () => {
+    this.setState(() => ({ show: false }));
```
jcaianirh: this.setState({ show: false }); should be good here.

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -61,10 +61,21 @@ class DashPluginCardView extends Component {
...
+  };
+  showModal = () => {
+    this.setState(() => ({ show: true }));
```
jcaianirh: this.setState({ show: false }); should be good here.

.eslintrc
jcaianirh: We use state= in other constructors, right? Why is this issue here?

jcaianirh: you need to move your statement up into the constructor. thats where you set up your state. the eslint was correct in flagging the error.

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
jcaianirh: also use format secondaryAction() {} for class methods.

gagansk: Changed the method format

.eslintrc
gagansk: Moved the state declaration to the constructor and removed the changes made to .eslintrc

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
gagansk: Comment has been addressed in the commit

gagansk:
@jcaianirh  comments have been addressed.

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -12,8 +12,8 @@ import {
   Card,
   CardHeading,
   DropdownKebab,
-  MenuItem,
   FieldLevelHelp,
+  MessageDialog,
```
jcaianirh: Within a dropdown kebab, we need to use menu items and not buttons. See http://patternfly-react.surge.sh/patternfly-3/index.html?knob-Title=Kebab%20title&selectedKind=patternfly-react%2FWidgets%2FDropdown%20Kebab&selectedStory=DropdownKebab&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

I would remove the Button, re add the Menu Item, and use that instead of the button.

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -92,15 +109,31 @@ class DashPluginCardView extends Component {
...
+        const primaryContent = <p className="lead">Are you sure?</p>;
+        const secondaryContent = <p>Plugin <b>{plugin.name}</b> will be permanently deleted</p>;
         return (
           <Col xs={12} sm={6} md={4} key={plugin.name}>
             <Card>
               <CardHeading>
                 <CardTitle>
                   <DropdownKebab id="myKebab" pullRight className="card-view-kebob">
-                    <MenuItem eventKey={plugin.name} onSelect={this.deletePlugin}>
+                    <Button onClick={this.showModal} bsStyle="primary">
```
jcaianirh: Add menu item here instead of button. See above comment.

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -59,17 +60,39 @@ class DashPluginCardView extends Component {
...
+      showConfirmation: true,
+      pluginToDelete: name,
+    });
+  }
+
 
   render() {
     let pluginCardBody;
     const { plugins } = this.props;
     const showEmptyState = isEmpty(plugins);
```
jcaianirh: if you use:
const { pluginToDelete, showConfirmation } = this.state;
you can eliminate this.state.pluginToDelete... whenever you use those state items. Same as what we do for props.

gagansk: @jcaianirh This has been addressed in the latest commit.

Issue #48, opened 2019-02-20
Upgrade to latest chris store api


jcaianirh:
Upgrade to @jbernal0019 's updated chris store api where the plugins use index as the key instead of name.

gagansk:
Closed with #52

PR #52, opened 2019-03-07
API Migration to 1.1.3 

gagansk:
This PR is to migrate from the ChRIS store API 1.1.2 to 1.1.3.
- Find and Delete Operations are altered.

- Get (Find) from Plugin List, Dashboard and Create Plugin page changed

- Delete Operation in Dashboard altered.


gagansk:
https://github.com/FNNDSC/ChRIS_store_ui/issues/48 Pull Request

gagansk:
@jcaianirh , thank you for your help in fixing the tests! I have updated the code. Please review.

gagansk:
@jcaianirh , Please don't approve the PR yet. With the new code, though the tests are passing, the feature seems to be broken. I will update soon.

gagansk:
Fixed the bug. Please review the changes, @jcaianirh

src/components/CreatePlugin/CreatePlugin.js
```diff
@@ -233,25 +233,31 @@ class CreatePlugin extends Component {
     const token = window.sessionStorage.getItem('AUTH_TOKEN');
     const client = new StoreClient(storeURL, { token });
 
-    let newPlugin;
+    let returnedPlugin;
```
jcaianirh: leave this newPlugin, we can use the setstate shorthand setState({newPlugin}).

src/components/CreatePlugin/CreatePlugin.js
```diff
@@ -233,25 +233,31 @@ class CreatePlugin extends Component {
...
+      returnedPlugin = await client.addPlugin(pluginName, pluginImage, file, pluginRepo);
     } catch ({ message }) {
       return this.handleError(message);
     }
 
     this.hideError();
     this.setState({ success: true });
-    return newPlugin;
+    this.setState({ newPlugin: returnedPlugin });
+    return returnedPlugin;
```
jcaianirh: no return statement needed here.

src/components/CreatePlugin/CreatePlugin.js
```diff
@@ -233,25 +233,31 @@ class CreatePlugin extends Component {
...
-      newPlugin = await client.addPlugin(pluginName, pluginImage, file, pluginRepo);
+      returnedPlugin = await client.addPlugin(pluginName, pluginImage, file, pluginRepo);
     } catch ({ message }) {
       return this.handleError(message);
     }
 
     this.hideError();
     this.setState({ success: true });
-    return newPlugin;
+    this.setState({ newPlugin: returnedPlugin });
```
jcaianirh: setState({newPlugin}) instead

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -94,7 +94,7 @@ class DashPluginCardView extends Component {
     const showEmptyState = isEmpty(plugins);
     const primaryContent = <p className="lead">Are you sure?</p>;
     const secondaryContent =
-      <p>Plugin <b>{pluginToDelete}</b> will be permanently deleted</p>;
+      <p>Plugin with ID <b>{pluginToDelete}</b> will be permanently deleted</p>;
```
jcaianirh: We want to show the plugin name here, not an id. The user doesn't care about the id #, they probably will never know what it is.

src/components/CreatePlugin/CreatePlugin.js
gagansk: eslint will produce the following error if the return statement is removed:
**error  Expected to return a value at the end of async method 'handleSubmit'**

So, the return statement will be unaltered.

src/components/CreatePlugin/CreatePlugin.js
gagansk: Comment is addressed in the commit.

src/components/CreatePlugin/CreatePlugin.js
gagansk: Comment is addressed in the commit.

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
gagansk: Now, the entire plugin will be sent to the function which handles Delete alert box. Another state variable pluginID, had to be introduced since ID is required to perform the delete operation and the Name is required to display to the user.

gagansk:
@jcaianirh, I have addressed the comments in the new commit.

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -72,17 +73,18 @@ class DashPluginCardView extends Component {
   }
   deletePlugin() {
     const { onDelete } = this.props;
-    const { pluginToDelete } = this.state;
-    onDelete(pluginToDelete);
+    const { pluginIDToDelete } = this.state;
+    onDelete(pluginIDToDelete);
```
jcaianirh: onDelete(pluginToDelete.id);

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -72,17 +73,18 @@ class DashPluginCardView extends Component {
...
   }
   secondaryAction() {
     this.setState({ showConfirmation: false });
   }
-  showModal(id) {
+  showModal(plugin) {
     this.setState({
       showConfirmation: true,
-      pluginToDelete: id,
+      pluginToDelete: plugin.name,
```
jcaianirh: For the state, just add the plugin, and call out the parameters you'll need in the delete method such as pluginToDelete.Id, pluginToDelete.Name, etc. Instead of having multiple state variables.

gagansk: If we add the entire plugin to the state, we will get an error since we will try to access the 'name' property of the plugin when the value is null in the beginning of the render() method in this line of code:
`<p>Plugin <b>{pluginToDelete}</b> will be permanently deleted</p>;`

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
gagansk: Comment has been addressed below

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
jcaianirh: @gagansk thats right, thats why we would need to change it to pluginToDelete.name

gagansk:
The DOM wouldn't render since it will try to access the 'name' property of the null value of the state variable pluginToDelete in the LOC: `<p>Plugin <b>{pluginToDelete.name}</b> will be permanently deleted</p>;`

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
gagansk: A conditional statement has been added to take care of the null condition.

PR #56, opened 2019-04-10
Graceful handling of Failure of Plugin Creation 

gagansk:
Create a conditional statement to handle the errors at the time of creation of plugin gracefully.

Current implementation: As of now, the entire response received from the server is begin rendered as it is in the UI.

Expected implementation: Display the reason for failing in creating the plugin.
Eg: A plugin with the given name already exists. or The URL format for the repo is not valid.

### DIFF
--- a/src/components/CreatePlugin/CreatePlugin.js
+++ b/src/components/CreatePlugin/CreatePlugin.js
@@ -83,7 +83,13 @@ class CreatePlugin extends Component {
   }
 
   handleError(message) {
-    this.setState({ formError: message });
+    if (message.includes('non_field_errors')) {
+      this.setState({ formError: 'A Plugin with this name already exists' });
+    } else if (message.includes('public_repo')) {
+      this.setState({ formError: 'Enter a valid URL for public repo.' });
+    } else {
+      this.setState({ formError: message });
+    }
   }
 
   hideError() {
--- a/src/components/CreatePlugin/CreatePlugin.js
+++ b/src/components/CreatePlugin/CreatePlugin.js
@@ -235,7 +235,8 @@ class CreatePlugin extends Component {
 
     let newPlugin;
     try {
-      newPlugin = await client.addPlugin(pluginName, pluginImage, file, pluginRepo);
+      const resp = await client.addPlugin(pluginName, pluginImage, file, pluginRepo);
+      newPlugin = resp.data;
     } catch ({ message }) {
       return this.handleError(message);
     }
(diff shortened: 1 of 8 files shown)


src/components/CreatePlugin/CreatePlugin.js
```diff
@@ -83,7 +83,16 @@ class CreatePlugin extends Component {
   }
 
   handleError(message) {
```
jcaianirh: @jbernal0019 whats the format of the error message? Is there a key we can look for to show specific messages? @gagansk this conditional looks for non_field_errors...could that be other errors besides the 'name already exists' error?

gagansk: I know two errors: 
Non-unique key for which the server response is 
{'non_field_errors': [ErrorDetail(string='The fields name, version must make a unique set.', code='unique')]}

Invalid repo URL for which the server response is 
{'public_repo': [ErrorDetail(string='Enter a valid URL.', code='invalid')]}

gagansk: @jcaianirh , these are the two errors which we are not handling and the UI displays the entire message response from the server as it is.

jcaianirh: @gagansk ok great, so we can key off the code and create a user friendly string for both cases and add a third default for some generic error message that we don't know about yet.

jcaianirh: @jbernal0019, could you reply and explain the error formatting above please? How did you envision the ui to handle the different types of errors it gets back from the js client.

jbernal0019: Hey @jcaianirh 

Those are raw server errors. Those are for client programmers like you. The UI will probably need to generate a nicer user-oriented error message that also makes sense when  considering the app context in which the error is generated.

jcaianirh: @jbernal0019 so the syntax is what? 
ErrorObject: {
  non_field_errors: [ {string: 'error string', code: 'error code'}, ...],
...
}

jbernal0019: Hey @jcaianirh @gagansk I've publish the latest version of the `chrisstoreapi` It's now `"1.1.7"`.

This new version formalizes better the API error handling. It basically follows the same error handling scheme as the Axios library. So given an error you first check if `error.response` exists. If so the error data is in `error.response.data`, otherwise you just use `error.message` to report the error.

Now `error.response.data` may be a plain error string or an object whose properties are the field names that produced the bad request. It may also have a property called "non_field_errors" when the error is not specifically related to a single field. The value of any of those properties is a list of plain string errors. This is the standard Django Rest Framework approach to reporting errors. 

Please upgrade the UI to use `chrisstoreapi "1.1.7"` and  let me know if you have any doubts.

PR #61, opened 2019-04-22
Server side validation of User creation

gagansk:
This PR has validation for the creation of users in store based on the server response.

### DIFF
--- a/src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
+++ b/src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
@@ -10,6 +10,7 @@ import {
   Button,
   Spinner,
 } from 'patternfly-react';
+import _ from 'lodash';
 import StoreClient from '@fnndsc/chrisstoreapi';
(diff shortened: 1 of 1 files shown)


src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
```diff
@@ -123,14 +123,29 @@ export class DeveloperSignup extends Component {
...
       } catch (e) {
-        /* TODO: JCC Enhance error handling */
-        this.setState({
-          loading: false,
-          error: {
-            message: 'This field must be unique',
-            controls: ['username'],
-          },
-        });
+        if (Object.prototype.hasOwnProperty.call(e, 'response')) {
```
jcaianirh: you can use lodash _.get here: https://lodash.com/docs/4.17.11#get

gagansk: Used _.has call: [https://lodash.com/docs/4.17.11#has]

PR #64, opened 2019-05-30
Fix text overflow

gagansk:
Fixing the issue of text overflow in the edit box committed in the previous pull request


gagansk:
@jcaianirh , I made a mistake in the previous PR. I had not saved the col size to 4 while committing for the previous PR because of which the edit text box was overflowing the Modal. I have corrected the mistake now. Please approve when you get time.

PR #100, opened 2020-10-11
fixed minor issue for plugins footer .

anujsingla:


mairin:
Looks great! I tested out here in Firefox and 👍

mairin (approved):
Tested this out in firefox and works well, and the method seems like a nice clean one. thanks!

Issue #108, opened 2020-11-13
Remove recompose dependency in favor of react hooks


jennydaman:
https://github.com/acdlite/recompose#a-note-from-the-author-acdlite-oct-25-2018

PR #109, opened 2020-11-13
Remove recompose

jennydaman:
To my knowledge, `compose()` called with only one argument does nothing.

There is only one use of `recompose` in the code which I have removed.

Solves #108, which I opened hastily without first looking at the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "react-dom": "16.4.1",
     "react-router-dom": "^4.3.1",
     "react-scripts": "3.4.3",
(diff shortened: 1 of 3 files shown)


anujsingla:
Reviewed.

anujsingla (approved):
Reviewed.

jennydaman:
Oh sorry I noticed that it says

> At least 1 approving review is required by **reviewers with write access**. 

I guess I can just merge these as admin instead of going by the rules...

Issue #13, opened 2018-07-23
Show failed network requests in the UI


jdtzmn:
Error messages should exist somewhere in the UI to inform the user if the page has trouble loading.

DivyanshiSingh:
Hey, @mairin I would like to work on this. Please assign me this issue.

jennydaman:
@mairin imo this kind of issue would seem to be something better to be worked on _after_ PatternFly4 rewrite

DivyanshiSingh:
Can you point me to the page where API fails and we don't get any notifications? @jennydaman

jennydaman:
First, let's find places in the code which are responsible for doing network requests.

Interaction with the API is made easy with the `@fnndsc/chrisstoreapi` npm package. Typical usage would be

```javascript
import Client from '@fnndsc/chrisstoreapi';

[...]

this.client = new Client(storeURL, auth);
```

And then something like

https://github.com/FNNDSC/ChRIS_store_ui/blob/0db51860bf76c3002785944edb13a62347194af2/src/components/Plugins/Plugins.js#L84-L91

To find all usages of this, we can search for the string `client.`
I will use the `grep` tool, you can use Github search, or your IDE's full-text search.
(A great alternative is [`rg`](https://github.com/BurntSushi/ripgrep))

```
$ grep -Fnr 'client.' src/
src/components/Dashboard/Dashboard.js:40:    return client.getPlugins(searchParams)
src/components/Dashboard/Dashboard.js:59:      response = client.getPlugin(pluginId).then(plugin => plugin.delete());
src/components/Dashboard/Dashboard.js:77:      response = client.getPlugin(pluginId)
src/components/Plugins/Plugins.js:85:      const star = await this.client.createPluginStar({ plugin_name: plugin.name });
src/components/Plugins/Plugins.js:100:      const star = await this.client.getPluginStar(previousStarState.id);
src/components/Plugins/Plugins.js:118:        plugins = await this.client.getPlugins(searchParams);
src/components/Plugins/Plugins.js:136:    const stars = await this.client.getPluginStars();
src/components/CreatePlugin/CreatePlugin.js:261:      const resp = await client.createPlugin(pluginData, fileObj);
src/components/Plugin/Plugin.js:72:      const star = await this.client.createPluginStar({ plugin_name: name });
src/components/Plugin/Plugin.js:86:      const star = await this.client.getPluginStar(previousStarState.id);
[...]

PR #123, opened 2021-04-07
Notifications for failed network requests

DivyanshiSingh:
Fixes: #13 
1) Created a reuseable Notification component to display the error messages.
2) Created an HttpApiCallError class to handle the errors.
3) Used try-catch in async functions.


DivyanshiSingh:
This is my initial implementation, please review it @jennydaman

src/errors/HttpApiCallError.js
```diff
@@ -0,0 +1,9 @@
+export default class HttpApiCallError extends Error {
+    constructor(message, statusCode) {
```
jennydaman, line 2: Please fix indentation here

src/components/Plugin/Plugin.js
```diff
@@ -41,12 +44,15 @@ export class Plugin extends Component {
...
-      pluginData = await this.fetchPluginData();
-    } else {
-      ({ pluginData } = this.state);
+    try{
+      if (!this.state.pluginData) {
+        pluginData = await this.fetchPluginData();
+      } else {
+        ({ pluginData } = this.state);
+      }
+    }catch(errors){
```
jennydaman, line 53: Future work would be to handle `errors` specifically.

src/components/Plugin/Plugin.js
```diff
@@ -41,12 +44,15 @@ export class Plugin extends Component {
...
+      this.showNotifications(new HttpApiCallError('Unable to fetch plugin data'));
     }
-
     if (this.isLoggedIn()) {
       this.fetchStarDataByPluginName(pluginData.name);
```
jennydaman, line 57: Put this in the `try` block too, it can fail.

src/components/Notifications/Notifications.js
```diff
@@ -0,0 +1,28 @@
...
+import React, { Component } from "react";
+import "./Notifications.css";
+class Notifications extends Component {
+  constructor(props) {
+    super(props);
+    this.deleteTime = 2000;
+  }
+  componentDidMount() {
+    this.timer = setInterval(() => {
+      console.log("here");
```
jennydaman, line 11: delete

src/components/Notifications/Notifications.js
```diff
@@ -0,0 +1,28 @@
...
+      this.props.closeNotification();
+    }, this.deleteTime);
+    console.log(this.props.message, this.props);
```
jennydaman, line 14: and that

jennydaman (commented):
This work is pretty amazing. A few coding things, like indentation and removal of print statements, have been commented upon.

Some feedback on style and notifications: The ChRIS_store_ui currently does not use many motion animations, so your notification feature stands out a bit. I think this could be alleviated by

- increasing the speed of the animation by 2-4x
- combine fade-in animation with slide-in animation
- change slide-in animation to move normal to the length of the notification box. i.e., have it slide in from the top instead of from the side.
- implementing an exit animation
- maybe have the exit animation be triggered by click, instead of on a timeout?

Error handling becomes a mess when you realize that there's a thousand ways something can go wrong for every way it can go right. Are you interested in adding the notification to more situations, or are you happy with what you have right now?

chris_store_ui_network_notification_1

DivyanshiSingh:
Ok, @jennydaman I will do all the required changes. 
Thank you for your feedback.

DivyanshiSingh:
> Some feedback on style and notifications: The ChRIS_store_ui currently does not use many motion animations, so your notification feature stands out a bit. I think this could be alleviated by
> * increasing the speed of the animation by 2-4x
   
> * combine fade-in animation with slide-in animation
> * change slide-in animation to move normal to the length of the notification box. i.e., have it slide in from the top instead of from the side.

 **I have increased the animation speed along with the fade-in animation**

> * implementing an exit animation
> * maybe have the exit animation be triggered by click, instead of on a timeout?

**I was planning to use the Alert component's props to give the functionality of closing notification on click, but as we are using an older version of patternfly-react , its not present there. I can implement it from scratch it will not be a very big change but we will have to rework once we change patternfly-react to a latest version.** How do you want me to proceed with this @jennydaman

jennydaman:
@DivyanshiSingh you're right, notification-style element is a pretty advanced feature and I agree that it would be better to hold off on a thorough implementation until we finish the migration to Patternfly4.

src/errors/HttpApiCallError.js
DivyanshiSingh: Resolved

src/components/Plugin/Plugin.js
DivyanshiSingh: Handled the failure inside the function fetchStarDataByPluginName

src/components/Plugin/Plugin.js
DivyanshiSingh: I think to handle the errors specifically I would need the structure of json response that we get. That will help me to design the HttpApiCallError according to the status code. Is this the right approach?

DivyanshiSingh: @jennydaman do we get structured error response from the API? I can use that to handle specific errors

jennydaman: Well, some responses are structured, but of course not every possibility can be covered.

In my original comment, I was originally thinking about handling errors we know of, like

- network timeout
- unexpected HTTP code (400, 500)
- cannot understand response

DivyanshiSingh: Ok got it. Will do the changes and let you know

jennydaman:
@DivyanshiSingh the imports are now in conflict, hmm...

If you would like to forward-port this feature to `pf4` using Patternfly4 components, that would be something great to have.

jennydaman (approved):
@DivyanshiSingh I have resolved a merge conflict with #134 

However since https://github.com/FNNDSC/ChRIS_store_ui/pull/144 was merged, maybe we should close this one. Is that okay with you, or do you want to merge this too?

DivyanshiSingh:
We have to close this.

Issue #53, opened 2019-03-13
Add Modify Plugin


jcaianirh:
Add the ability to modify a plugin once its created. Use @jbernal0019 's backend js api for modifiable fields.

zrthxn:
@mairin I could start working on this if I can get some documentation on the backend API. Also please let me know if there are specific requirements for this

jennydaman:
@zrthxn some documentation is here: https://fnndsc.github.io/ChRIS_store/

But more helpful is the API itself. It's designed to be self-documenting in a way. (does it work? you be the judge)

For example: 


```
https https://chrisstore.co/api/v1/20/
HTTP/1.1 200 OK
Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
Connection: Keep-Alive
Content-Length: 973
Content-Type: application/vnd.collection+json
Date: Thu, 08 Apr 2021 06:54:39 GMT
Keep-Alive: timeout=5, max=100
Server: Apache
Vary: Accept,Cookie
X-Frame-Options: SAMEORIGIN

{
    "collection": {
        "href": "https://chrisstore.co/api/v1/20/",
        "items": [
            {
                "data": [
                    {
                        "name": "id",
                        "value": 20
                    },
                    {
                        "name": "creation_date",
                        "value": "2021-02-08T11:49:15.818028-05:00"
                    },
                    {
                        "name": "modification_date",
                        "value": "2021-02-09T07:44:00.312420-05:00"
                    },
                    {
                        "name": "name",
                        "value": "pl-med2img"
                    },
                    {
                        "name": "title",
                        "value": "A ChRIS plugin app"
                    },
                    {
                        "name": "stars",
                        "value": 0
                    },
                    {
                        "name": "public_repo",
                        "value": "https://github.com/FNNDSC/pl-med2img"
                    },
                    {
                        "name": "license",
                        "value": "MIT"
                    },
                    {
                        "name": "type",
                        "value": "ds"
                    },
                    {
                        "name": "icon",
                        "value": ""
                    },
                    {
                        "name": "category",
                        "value": ""
                    },
                    {
                        "name": "authors",
                        "value": "Arushi Vyas <dev@babyMRI.org>"
                    },
                    {
                        "name": "documentation",
                        "value": "https://github.com/FNNDSC/pl-med2img"
                    }
                ],
                "href": "https://chrisstore.co/api/v1/20/",
                "links": [
                    {
[...]

PR #126, opened 2021-04-09
Fix runtime error on opening Dashboard

zrthxn:
Fixed the following issues
1. **TypeError: DashApplicationType is not a constructor** on opening Dashboard after creating a plugin, described in detail [here.](https://github.com/FNNDSC/ChRIS_store_ui/issues/53#issuecomment-816682408)
2. Changed change public repo  Link in PluginBody to anchor tag to open the link instead of navigating to it as a path.


src/components/Plugin/components/PluginBody/PluginBody.js
```diff
@@ -137,12 +137,12 @@ const PluginBody = ({ pluginData }) => (
                         </div>
                           <div className="plugin-body-public-repo">
                             <h4>Public Repo:</h4>
-                            <Link
+                            <a
                               href={pluginData.public_repo}
                               to={pluginData.public_repo}
```
jennydaman, line 142: This is a good catch. But since we're using `<a>`, should we delete the `to=`?

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -173,7 +173,7 @@ class DashPluginCardView extends Component {
     if (plugins) {
       pluginCardBody = plugins.map((plugin) => {
         const creationDate = new RelativeDate(plugin.creation_date);
-        const applicationType = new DashApplicationType(plugin.type);
+        const applicationType = DashApplicationType(plugin.type);
```
jennydaman, line 176: Thanks!

src/components/Plugin/components/PluginBody/PluginBody.js
zrthxn: Yes we should I think I must've overlooked that

jennydaman (approved):

Issue #101, opened 2020-10-15
TODO: Upgrade ChRIS Store from PF3 to PF4


mairin:

mindreeper2420:
Looking at the last time the `package.json` file was updated, PatternFly Core and PatternFly React are around 2 years old.

The last version of PatternFly React 2 `patternfly-react v2.8.0` was released around May 2018.  The last version of PatternFly 3 was last released in November of 2018).

@mairin looking at the documentation from PatternFly, as well as the fact that everything is quite old (and hard to upgrade from), what would you like to do to move forward?

mairin:
@mindreeper2420 what do you think? is it better to rewrite to 4?

mindreeper2420:
@mairin IMO, I'd rewrite in 4 (it would open up the UI to newer features/designs, cleaner code, flexibility, etc). PF3 is not supported anymore, so there is also that.

You can run PatternFly 3 and PatternFly 4 together, but that does increase the site size. I've tested out the PF3/PF4 combo and it does run without any noticeable issues, when no changes are applied to the UI. The issues arise when you start to add high-level layout changes (such as to the content wrapper or grid layout), but updating component-level pieces (such as buttons and links) are looking okay.

I've attached a few screenshots - one with high-level changes and one with individual component-level changes.

**Page-level changes (content wrapper+cascading styles, buttons, links, lists)**
![Screen Shot 2020-11-09 at 10 55 10 AM](https://user-images.githubusercontent.com/4032718/98564186-0c246b80-227a-11eb-8dc6-76b7ea9d87e0.png)


**Component-level changes (headers, buttons, links, list)**
![Screen Shot 2020-11-09 at 10 50 09 AM](https://user-images.githubusercontent.com/4032718/98563590-58bb7700-2279-11eb-8804-b12564960b7f.png)

mindreeper2420:
@mairin here is a branch that I tested this on, if you'd like to take a look: https://github.com/mindreeper2420/ChRIS_store_ui/tree/upgrade-chris-store-ui

jennydaman:
I have been trying to work on this issue but it's dependency hell. `@patternfly/react-core@4.75.2` does not work with `react@16.4.1`, bumping to `react@^16.8.0` breaks tests.

However it's not entirely fair that the tests are failing, they are producing errors on Travis as well (yet without failing for some reason)

https://travis-ci.com/github/FNNDSC/ChRIS_store_ui/builds/201177671

```
[...]

PR #127, opened 2021-04-09
Search Component enhancement and Fix for small screens

DivyanshiSingh:
Enhancement: #101 
1. Implemented a configurable search component
2. Brought search component to the horizontal nav in smaller screens
3. Changed functionality of search icon: User can click on the icon to search

s1
s2
s3

### DIFF
--- a/src/components/Navbar/Navbar.js
+++ b/src/components/Navbar/Navbar.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 import {
   Brand,
   PageHeader,
@@ -35,6 +35,7 @@ const navLinks = [
  * Conditionally renders a list of links into a <Nav>.
  */
 const Navigation = ({ store }) => {
+  const [searchKey, setSearchKey] = useState('');
   const shouldShowLink = (linkInfo) => {
     if (!linkInfo.cond) {
       return true;
@@ -58,6 +59,15 @@ const Navigation = ({ store }) => {
               </NavItem>
             ))
         }
+        <NavItem>
+          <Search 
+            placeholder="Search Plugin"
+            value={searchKey}
+            onChange={(value) => setSearchKey(value)}
+            onClear={() => setSearchKey('')}
+            onSearch={() => {console.log('Enter pressed')}}
+          />
+        </NavItem>
       </NavList>
     </Nav>
   );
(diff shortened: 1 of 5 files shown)


DivyanshiSingh:
@jennydaman have a look at this. I was thinking why are we not using react-router-dom v5? (this will provide us the functionality to switch between routes using useHistory() hooks which is not available in 4.3.1)

src/components/Navbar/Navbar.js
```diff
@@ -35,6 +35,7 @@ const navLinks = [
  * Conditionally renders a list of links into a <Nav>.
  */
 const Navigation = ({ store }) => {
+  const [searchKey, setSearchKey] = useState('');
```
jennydaman, line 38: yay, hooks!

src/components/Navbar/Navbar.js
```diff
@@ -80,9 +90,6 @@ const StatefulLoginButton = ChrisStore.withStore(LoginButton);
 
 const HeaderTools = (
   <PageHeaderTools>
-    <PageHeaderToolsItem>
```
jennydaman, line 83: I put `<Search />` in `<PageHeaderToolsItem>` for the sake of using `<PageHeaderToolsItem>`. Perhaps for better "integration" with Patternfly's design opinions. I assume the reason you are changing this is for greater flexibility with placement?

jennydaman:
The screenshots were much appreciated.

Maybe we should collapse the word "Seach" and show the symbolic icon only on small screens.

I haven't looked into react-router-dom v5. I'd say go for it. Aside from Patternfly 3 --> 4, another objective of this issue is to modernize and update all dependencies. It can get messy because incompatibilities happen every time.

src/components/Navbar/Navbar.js
DivyanshiSingh: Yes the placement was the reason to change the position of Search Component

DivyanshiSingh:
I have changed the react-router-dom version and the Search icon changes you asked. Please have a look.

jennydaman:
I see that you've removed the text. Iconography should speak for itself, good thinking.

screenshot

Once expanded, the icon gets misplaced.

btw, the target branch https://github.com/FNNDSC/ChRIS_store_ui/tree/pf4 itself is a development branch, Changes can be merged into it whenever you feel like, so let me know.

DivyanshiSingh:
> Once expanded, the icon gets misplaced.
I have fixed this, It was a css issue.

jennydaman:
Please create new PRs for subsequent features, such as for `Button`

jennydaman:
Hmm, on wide screens the search bar's left-justify looks weird. Maybe have the bar expand to take up all the space it can have, or restore the right-justified design.

screenshot

Issue #129, opened 2021-04-10
Correct the placeholder text on chrisstore.co/create


jennydaman:
![screenshot](https://user-images.githubusercontent.com/20404439/114259364-634e7980-999b-11eb-942f-13453d795963.png)

Where it says

> :star: 10k+ :clock10: Last modified 1 hour ago

archanaserver:
Hi @jennydaman I want to work on this issue.

PR #130, opened 2021-04-10
fix: placeholder text on chrisstore.co/create

archanaserver:
Fixes: [#129](https://github.com/FNNDSC/ChRIS_store_ui/issues/129)

@jennydaman please review this. I have fixed the placeholder text on chrisstore.co/create.


jennydaman (approved):
LGTM. Though, in review I've noticed a strangely sized off-center element. Probably not the fault of this PR, but something to note as it might cause a problem in the future.

screenshot

archanaserver:
Oh yes I see, I'll look into that too. @jennydaman

Issue #7, opened 2018-07-16
Fuzzy queries must be possible for plugin's name.


jdtzmn:
See FNNDSC/ChRIS_store#2.

gagansk:
Can the issue be described more? Is it Fuzzy queries must be supported for plugin search or Fuzzy queries must be used to suggest the plugin name dynamically based on the availability during creation?

mairin:
I think this is for searching for plugins. I would say the first task here would be to get the ChRIS store up and running, add a bunch of new plugins, and then test out the search to see how exact queries must be.

sonylomo:
Hey @mairin, I'd like to work on this 😁.

sonylomo:
Is it ok to use a library like Fuse.js to implement fuzzy search?

jennydaman:
Fuzzy search is better implemented server-side (faster).

The back-end supports partial matching of queries, for example `https://chrisstore.co/api/v1/plugins/search?name=pl-fetal` returns `pl-fetal-brain-mask` and `pl-fetal-brain-assessment`. More advanced fuzzy querying, such as typo correction, might be out of scope — especially because names might use esoteric abbreviations such as `pl-pfdo_med2img`.

I suppose the real issue for now is to correctly implement search against the `/api/v1/plugins/search?name=%s` endpoint.

Another related goal would be to implement [OpenSearch](https://developer.mozilla.org/en-US/docs/Web/OpenSearch). I've created a separate issue for this https://github.com/FNNDSC/ChRIS_store_ui/issues/121

PR #134, opened 2021-04-11
Search query fix

sonylomo:
Fixes issue #7


src/components/Plugins/Plugins.js
```diff
@@ -1,15 +1,14 @@
-import React, { Component } from 'react';
```
jennydaman, line 1: hmm, I am surprised that `eslint` does not care about single vs double quote marks.

jennydaman (approved):
Search was such an important feature, thank you for the work!

Issue #137, opened 2021-04-11
List plugins by category


jennydaman:
![screenshot](https://user-images.githubusercontent.com/20404439/114314038-1f589300-9ac7-11eb-8755-8fcc69295ea2.png)

The left sidebar showing Categories is currently filler.

The backend can be queried for plugins belonging to a certain category.

```
https://chrisstore.co/api/v1/plugins/search/?category=Segmentation
```

Considerations: can the backend be queried for a list of categories?
For the time being, it might be better to hard-code the list of categories into the front-end.

zrthxn:
I'd like to start working on this.

zrthxn:
I wanted to ask, since you mentioned in another issue that you were looking to move towards functional components, would it be okay if parts of the plugin list component were ported to functional components? Since filtering would be a new feature to add, maybe it'd make sense to revamp the component along the way?

jennydaman:
That was mentioned by @PintoGideon in a different repo: https://github.com/FNNDSC/ChRIS_ui/issues/206

I have used the word "functional" in passing, meaning "working," not the JSX jargon. Though I suppose it'd be an upgrade.

Regarding revamping, the effort is happening in the issue https://github.com/FNNDSC/ChRIS_store_ui/issues/101 and being spearheaded by @DivyanshiSingh. If you are concerned about refactors, beautiful code, and doing things for the long run, consider making PRs to the `pf4` branch instead.

zrthxn:
I've found that instead of querying the backend for plugins in a category its simpler to filter by category on the frontend since I'm still not sure if the backend can be queried for a list of categories (after reading the docs for a while).

For the list of categories, I'm accumulating the list from the plugins which are fetched when the page loads.

jennydaman:
@zrthxn In the first comment, I gave an example for how to query the backend API for plugins belonging to a given category. The example is reproduced below:

```
https://chrisstore.co/api/v1/plugins/search/?category=Segmentation
```

Reading the docs is usually not productive because, well, there's not much. It is shameful to say this, but the best way to understand the API is to:

1. ask me
2. try things out

As I've explained in https://github.com/FNNDSC/ChRIS_store_ui/issues/53#issuecomment-815506288, the API is self-documenting in some ways and it always links to itself.

In general, if something can be done server-side, it's better to do it server-side. Reasons:

[...]

PR #138, opened 2021-04-11
Add Plugin filtering by Category

zrthxn:
Added functionality to filter list of plugins by clicking on a category.

**Notes**
- The backend is not queried for list of categories. 
- The list of categories is accumulated from the plugins that are fetched when the page loads.

Screenshot (36)

This is still an initial implementation. Edits can be made according to comments and reviews.
Partially implements #137


zrthxn:
The build is failing because there is a test to check if there are only 3 categories while we have more than that.

src/components/Plugins/Plugins.js
```diff
@@ -43,14 +43,34 @@ export class Plugins extends Component {
...
+    this.handleCategorySelect = this.handleCategorySelect.bind(this)
   }
 
   componentWillMount() {
     this.mounted = true;
   }
 
   componentDidMount() {
-    this.fetchPlugins().catch((err) => {
+    this.fetchPlugins().then((plugins) => {
```
jennydaman, line 54: As we're visiting the code, we should refactor what we can to use `async`/`await`

jennydaman (commented):
Don't worry about travis builds much. If you break any tests locally, it's ok to disable them for the time being. Just leave a comment. Or better, correct them.

src/components/Plugins/Plugins.js
zrthxn: Will do.

src/components/Plugins/Plugins.js
```diff
@@ -43,14 +43,34 @@ export class Plugins extends Component {
...
+      /**
+       * Temporary
+       * Accumulate categories from fetched plugins
+       */
+      let { categories } = this.state, list = [];
```
jennydaman, line 59: I have mentioned this issue in https://github.com/FNNDSC/ChRIS_store_ui/issues/137#issue-855358066

Instead of trying to discover categories, in this case it would be better to hard-code them.

So this code isn't necessary, but there's more to discuss for the sake of interest.

You're trying to solve a very common problem: you want a "collection" of things. The collection is *unordered* and _does not contain duplicates_. There exists a data structure, called a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), which handles these features/constraints more easily and efficiently than a list. A set typically runs in O(1) or O(logn) whereas your solution, using a nested loop (map is a loop, reduce is a loop), has O(n^2) efficiency.

src/components/Plugins/Plugins.js
```diff
@@ -43,14 +43,34 @@ export class Plugins extends Component {
...
+      plugins.reduce((_, current) => {
+        // if list doesn't include category, add it
+        if (current.category && !list.includes(current.category)) {
+          list.push(current.category)
+          categories.push({
+            name: current.category, length: 1
+          });
+        // Else increment count
+        } else if (list.includes(current.category)) {
+          categories[categories.map(({ name }, _) => name).indexOf(current.category)].length++;
```
jennydaman, line 69: Also it's no coincidence that the list processing function is called "map." There is another data structure which would handle this generic problem more efficiently too, it's [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

Plain objects work as well, though `Map` is a more efficient implementation.

jennydaman (changes requested):
I've left comments for computer science nerds.
The changes I am requesting are:

- [ ] hard-code the category names
- [ ] use backend API to get plugins belonging to each hard-coded category

PR #139, opened 2021-04-11
Add Plugin filtering by Category

zrthxn:
Added functionality to filter list of plugins by clicking on a category.

**Notes**
- [x] hard-coded the category names
- [x] used backend API to get plugins belonging to each hard-coded category
- [x] The count of categories is accumulated from the plugins that are fetched when the page loads.

Screenshot (36)

Partially implements #137


zrthxn:
@jennydaman Made a new PR with the requested changes.

src/components/Plugins/Plugins.js
```diff
@@ -27,32 +27,42 @@ export class Plugins extends Component {
...
-        },
+        { name: "Visualization", length: 3 },
+        { name: "Modeling", length: 11 },
+        { name: "Statistical Operation", length: 7 },
+        { name: "FreeSurfer", length: 0 },
+        { name: "MRI Processing", length: 0 },
       ],
     };
 
     this.fetchPlugins = this.fetchPlugins.bind(this);
```
jennydaman, line 38: Might be able to use some sugar here.
https://reactjs.org/docs/faq-functions.html

jennydaman (changes requested):

src/components/Plugins/Plugins.js
```diff
@@ -27,32 +27,42 @@ export class Plugins extends Component {
...
-        },
-        {
-          name: "Modeling",
-          length: 11,
-        },
-        {
-          name: "Statistical Operation",
-          length: 7,
-        },
+        { name: "Visualization", length: 3 },
```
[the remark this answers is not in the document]
zrthxn: They'll stay at zero then since no current plugin has those categories. I just thought that'd look bad.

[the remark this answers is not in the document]
jennydaman: It's the truth ¯\_(ツ)_/¯

src/components/Plugins/Plugins.js
zrthxn: Do you mean something like [this?](
https://reactjs.org/docs/faq-functions.html#class-properties-stage-3-proposal)
```js
fetchPlugins = () => {
  // fetch
}
```

jennydaman: Yes if supported

src/components/Plugins/Plugins.js
```diff
@@ -27,32 +27,39 @@ export class Plugins extends Component {
...
-        },
-        {
-          name: "Modeling",
-          length: 11,
-        },
-        {
-          name: "Statistical Operation",
-          length: 7,
-        },
+        { name: "Visualization", length: 0 },
```
jennydaman, line 30: Is there a reason why you do not use a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) or regular object instead of an array of entries?

`Map` would be more efficient and greatly simplify the moving parts of `componentDidMount()`

jennydaman: I assume, to keep the rest of code as it is... OK if you are aiming for a minimal patchset.

zrthxn: It was mainly to maintain compatibility with the PluginCategories component. Changing to a Map makes that one part easier but introduces more complications elsewhere.

jennydaman (approved):

PR #142, opened 2021-04-12
Button Component: pf4

DivyanshiSingh:
Fixes: #101 
1. Created a generalised Button component to be used throughout the project.
2. Replaced all the occurrences of Button (react-patternfly) to use v4 of patternfly

### DIFF
--- a/src/components/Button/button.css
+++ b/src/components/Button/button.css
@@ -1,0 +1,7 @@
+@media (max-width: 480px){
+.chris-button button{
+    width:70px;
+    height:auto;
+    font-size:0.9em;
+}
+}
--- a/src/components/Button/button.css
+++ b/src/components/Button/button.css
@@ -1,7 +1,3 @@
-@media (max-width: 480px){
-.chris-button button{
-    width:70px;
-    height:auto;
-    font-size:0.9em;
-}
+.other-button{
+    width: 100%;
 }
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -1,0 +1,35 @@
+import React from 'react';
+import { Button } from '@patternfly/react-core';
+import PropTypes from 'prop-types';
+const ButtonComponent = (props) => {
+  const {
+    variant,
+    onClick,
+    customClass,
+    icon,
+    loading,
+  } = props;
+  return(
+    <div className="chris-button">
+      <Button
+        isLoading={loading}
+        spinnerAriaValueText
+        variant={variant}
+        onClick={onClick}
+        className={customClass}
+        {...props}
+      >
+        {props.children}
+        </Button>
+    </div>
+  );
+};
+ButtonComponent.propTypes = {
+  variant: PropTypes.string,
+  onClick: PropTypes.func,
+  icon: PropTypes.object,
+  loading: PropTypes.bool,
+  customClass: PropTypes.string
+};
+
+export default ButtonComponent;
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -1,30 +1,30 @@
-import React from 'react';
-import { Button } from '@patternfly/react-core';
-import PropTypes from 'prop-types';
-import { useHistory } from 'react-router-dom';
-const ButtonComponent = (props) => {
-  const {
-    variant,
-    onClick,
-    customClass,
-    loading,
-    toRoute,
-    children,
-    ...otherProps
-  } = props;
+import React from "react";
+import { Button } from "@patternfly/react-core";
+import PropTypes from "prop-types";
+import { useHistory } from "react-router-dom";
+import "./button.css";
+const ButtonComponent = ({
+  variant,
+  onClick,
+  customClass,
+  loading,
+  toRoute,
+  children,
+  ...otherProps
+}) => {
   const history = useHistory();
-  return(
+  return (
     <div className="chris-button">
       <Button
         isLoading={loading}
         spinnerAriaValueText
         variant={variant}
-        onClick={toRoute?history.push(toRoute):onClick}
+        onClick={toRoute ? history.push(toRoute) : onClick}
         className={customClass}
         {...otherProps}
       >
         {children}
-        </Button>
+      </Button>
     </div>
   );
 };
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -6,7 +6,6 @@ import "./button.css";
 const ButtonComponent = ({
   variant,
   onClick,
-  customClass,
   loading,
   toRoute,
   children,
@@ -14,14 +13,12 @@ const ButtonComponent = ({
 }) => {
   const history = useHistory();
   return (
-    <div className="chris-button">
+    <div  {...otherProps}>
       <Button
         isLoading={loading}
-        spinnerAriaValueText
         variant={variant}
         onClick={toRoute ? history.push(toRoute) : onClick}
-        className={customClass}
-        {...otherProps}
+        className="other-button"
       >
         {children}
       </Button>
@@ -33,7 +30,6 @@ ButtonComponent.propTypes = {
   onClick: PropTypes.func,
   icon: PropTypes.object,
   loading: PropTypes.bool,
-  customClass: PropTypes.string
-};
+  };
 
 export default ButtonComponent;
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -1,21 +1,23 @@
 import React from 'react';
 import { Button } from '@patternfly/react-core';
 import PropTypes from 'prop-types';
+import { useHistory } from 'react-router-dom';
 const ButtonComponent = (props) => {
   const {
     variant,
     onClick,
     customClass,
-    icon,
     loading,
+    toRoute,
   } = props;
+  const history = useHistory();
   return(
     <div className="chris-button">
       <Button
         isLoading={loading}
         spinnerAriaValueText
         variant={variant}
-        onClick={onClick}
+        onClick={toRoute?history.push(toRoute):onClick}
         className={customClass}
         {...props}
       >
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -9,6 +9,8 @@ const ButtonComponent = (props) => {
     customClass,
     loading,
     toRoute,
+    children,
+    ...otherProps
   } = props;
   const history = useHistory();
   return(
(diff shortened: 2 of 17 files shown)


DivyanshiSingh:
@jennydaman please review this, next I am taking up the notifications component as you asked it to be implemented using pf4.

src/components/Button/index.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Button } from '@patternfly/react-core';
+import PropTypes from 'prop-types';
+const ButtonComponent = (props) => {
+  const {
+    variant,
+    onClick,
+    customClass,
+    icon,
```
jennydaman, line 9: unused `icon`, is this something you intended to implement but found no usages?

jennydaman (commented):
The `history` object is being passed down from a root router element. Maybe it's possible to import `useHistory` into `<Button />` instead?

https://reactrouter.com/web/api/Hooks/usehistory

(not sure, I just found this from a quick search --- lmk what you think)

src/components/Button/index.js
DivyanshiSingh: Yes, I thought of Icon Buttons but found no usages, I will be removing it for now

DivyanshiSingh:
I have added history in Button component and we can route using the Buttons 'toRoute' prop

src/components/Welcome/components/WelcomeChRIS/WelcomeChRIS.js
```diff
@@ -43,18 +44,18 @@ const WelcomeChRIS = () => (
     <div className="row welcome-chris-btn-row">
       <div>
         <Button
-          bsStyle="default"
-          bsSize="large"
-          className="welcome-chris-btn welcome-chris-more-btn"
+          variant="secondary"
+          customClass="welcome-chris-btn welcome-chris-more-btn"
+          onClick={() => {}}
```
jennydaman, line 49: is a no-op required?

src/components/Button/index.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Button } from '@patternfly/react-core';
+import PropTypes from 'prop-types';
+import { useHistory } from 'react-router-dom';
+const ButtonComponent = (props) => {
```
jennydaman, line 5: I think spread is being misused here.

You should spread the arguments:

```javascript
const ButtonComponent = ({variant, onClick, customClass, loading, toRoute, ...props}) => {
```

Syntax is a bit cleaner and fixes a problem where later on line 22 you have

```javascript
{...props}
```

Without spreading out `children` from `props`, I think you're accidentally putting `childen` into both the attributes and children of `<Button />`

jennydaman (changes requested):

src/components/Welcome/components/WelcomeChRIS/WelcomeChRIS.js
DivyanshiSingh: I don't know whats the function of this button

src/components/Button/index.js
DivyanshiSingh: I have made the required changes.

src/components/Welcome/components/WelcomeChRIS/WelcomeChRIS.js
jennydaman: The button is a placeholder, though I am asking about

```javascript
          onClick={() => {}}
```

Is it necessary to have this everywhere?

DivyanshiSingh: No we don't need it. I have removed it.

src/components/Button/index.js
jennydaman: Hmm, I do not see the suggestion implemented.

Besides, what is the point of wrapping Patternfly's `<Button />`? It doesn't seem like the wrapper is doing much. Perhaps to provide the `toRoute` prop but I think importing `useHistory` is easy enough. But that can be up to you. Also the class `chris-button` is undefined.

DivyanshiSingh: ooh sorry for this @jennydaman. I think I missed some commits let me check.

DivyanshiSingh: I wrapped the Button so that we can customize it if needed, but for now I can only see that toRoute is the only customization. Should I remove the wrapper @jennydaman ?

jennydaman: Hmm,on second thought it might be worth keeping. Typically I would try to use the UI components as-is (as how the library developer intends). But customization might be something worth pursuing in the future. Since the ChRIS store's purpose is to be an advertisement, developing our own branding (hence deviating from Patternfly) is a good thing.

src/components/Button/button.css
```diff
@@ -0,0 +1,7 @@
+@media (max-width: 480px){
+.chris-button button{
+    width:70px;
```
jennydaman, line 3: Is a fixed-width something we want to be setting globally to all `<Button />`?
This might make it hard to customize widths of buttons.

src/components/Button/index.js
```diff
@@ -1,28 +1,30 @@
...
     <div className="chris-button">
       <Button
         isLoading={loading}
         spinnerAriaValueText
         variant={variant}
-        onClick={toRoute?history.push(toRoute):onClick}
+        onClick={toRoute ? history.push(toRoute) : onClick}
         className={customClass}
-        {...props}
+        {...otherProps}
```
jennydaman, line 24: Typically, `otherProps` are given to the parent element returned by `render`. Otherwise it's impossible to customize the parent.

src/components/Button/button.css
```diff
@@ -0,0 +1,7 @@
+@media (max-width: 480px){
+.chris-button button{
```
jennydaman, line 2: I think it would be better to give the `other-button` class to `<Button />` instead of using a child selector.

src/components/Button/index.js
```diff
@@ -1,28 +1,30 @@
...
+  children,
+  ...otherProps
+}) => {
   const history = useHistory();
-  return(
+  return (
     <div className="chris-button">
       <Button
         isLoading={loading}
         spinnerAriaValueText
```
jennydaman, line 20: Remove undefined attribute

src/components/Button/index.js
```diff
@@ -1,28 +1,30 @@
...
-  } = props;
+import React from "react";
+import { Button } from "@patternfly/react-core";
+import PropTypes from "prop-types";
+import { useHistory } from "react-router-dom";
+import "./button.css";
+const ButtonComponent = ({
+  variant,
+  onClick,
+  customClass,
```
jennydaman, line 9: `customClass` seems like an awkward workaround.
Perhaps a better solution would be to give the child `<Button />` a class (if necessary) and then spread `otherProps` as attributes to the parent `<div>` (so if `otherProps` contains `className`, it will be applied to the parent)

src/components/Button/button.css
DivyanshiSingh: Removed this and gave other-button class to <Button/>

src/components/Button/button.css
DivyanshiSingh: done

src/components/Button/index.js
DivyanshiSingh: implemented you suggestions to use otherProps with root element to expose it for customization

src/components/Button/index.js
DivyanshiSingh: done

src/components/Button/index.js
DivyanshiSingh: changed the approach to make the root element as customizable

src/components/Navbar/Navbar.css
```diff
@@ -216,7 +216,10 @@ header > .navbar-dropdown {
   /* font */
   font-size: 1.5em;
 }
-
+.login-button .other-button{
```
jennydaman, line 219: Might be able to set colors without using `!important` using the CSS variables as per documentation.
https://www.patternfly.org/v4/components/button/#css-variables

Also I am hesitant about referencing the `other-button` class here. It's fine for now, but I've created a new issue proposing a fix: https://github.com/FNNDSC/ChRIS_store_ui/issues/151

src/components/Button/index.js
```diff
@@ -0,0 +1,35 @@
...
+  variant,
+  onClick,
+  loading,
+  toRoute,
+  children,
+  ...otherProps
+}) => {
+  const history = useHistory();
+  return (
+    <div  {...otherProps}>
```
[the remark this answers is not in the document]
DivyanshiSingh: I am using a parent wrapper so that in the future if we want an icon button or a loader button both being different from what is provided by patternfly by default, we can add those elements in the parent wrapper along with Button. What do you think about this @jennydaman?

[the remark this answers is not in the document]
jennydaman: OK, let's keep in then.

I think loading and icon buttons are both natively supported by PatternFly.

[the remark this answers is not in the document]
DivyanshiSingh: Yes but let's say we want a different loader and loader behaviour, I was thinking about that

src/components/Navbar/Navbar.css
DivyanshiSingh: I am unable to change the color using css-variables as i think we can't use button variables in Navbar. Also, !important is required to override the color that the Navbar is providing

jennydaman: Hmm, I was a bit fond of the dark button.
original
Your blue seems a bit too striking, though it's ok too. Paging @mairin for a professional opinion.
changed

BTW you can use CSS variables here.

```css
.login-button .pf-c-button.pf-m-primary {
  --pf-c-button--m-primary--BackgroundColor: red;
}
```

result

DivyanshiSingh: so should I remove the blue?

mairin: It's the primary call to action so i would keep it filled in blue. If you can shift the blue color to the same blue color that is the bg color of the jumbo below (that is ChRIS blue) it would be better, but the PatternFly default blue is ok.

@jennydaman - the darker button is aesthetically pleasing but doesn't draw the eye as much as it needs to IMHO. when in doubt i think sticking with the PF standard practice is better. I would classify this as a primary button which is why I'd say fill it.

DivyanshiSingh:
@jennydaman if the color is ok for now, can you merge this PR so that I can use the responsive button in my Autocomplete branch?

jennydaman (approved):

zrthxn:
The `useHistory` hook seems to cause this issue on the latest version pulled from the pf4 branch.

Screenshot (50)

DivyanshiSingh:
@zrthxn run ```npm i``` . You must be using older version of reat-router-dom.

zrthxn:
I have version 5.2.0
Still I'll try

PR #144, opened 2021-04-13
UI Notifications for errors

DivyanshiSingh:
Fixes: #13 and #101 
1) Created a reuseable notification component in patternfly-4 to display the error messages.

### DIFF
--- a/src/components/Dashboard/Dashboard.js
+++ b/src/components/Dashboard/Dashboard.js
@@ -66,7 +66,7 @@ class Dashboard extends Component {
     let response;
     try {
       response = await client.getPlugin(pluginId);
-      response.delete();
+      await response.delete();
       if(response.data) {
         this.fetchPlugins();
       }
--- a/src/components/Dashboard/Dashboard.js
+++ b/src/components/Dashboard/Dashboard.js
@@ -7,6 +7,8 @@ import DashPluginCardView from './components/DashPluginCardView/DashPluginCardVi
 import DashTeamView from './components/DashTeamView/DashTeamView';
 import DashGitHubView from './components/DashGitHubView/DashGitHubView';
 import ChrisStore from '../../store/ChrisStore';
+import Notification from '../Notification';
+import HttpApiCallError from '../../errors/HttpApiCallError';
 
 class Dashboard extends Component {
   constructor(props) {
@@ -14,6 +16,7 @@ class Dashboard extends Component {
     this.state = {
       pluginList: null,
       loading: true,
+      error: null,
     };
     this.initialize = this.initialize.bind(this);
     this.deletePlugin = this.deletePlugin.bind(this);
@@ -22,10 +25,15 @@ class Dashboard extends Component {
 
   componentDidMount() {
     this.fetchPlugins().catch((err) => {
+      this.showNotifications(new HttpApiCallError(err));
       console.error(err);
     });
   }
-
+  showNotifications = (error) => {
+    this.setState({
+      error: error.message,
+    })
+  }
   fetchPlugins() {
     const { store } = this.props;
     const storeURL = process.env.REACT_APP_STORE_URL;
@@ -61,6 +69,7 @@ class Dashboard extends Component {
         this.fetchPlugins();
       });
     } catch (e) {
+      this.showNotifications(new HttpApiCallError(e));
       return e;
     }
     return response;
@@ -81,6 +90,7 @@ class Dashboard extends Component {
         this.fetchPlugins();
       });
     } catch (e) {
+      this.showNotifications(new HttpApiCallError(e));
       return e;
     }
     return response;
@@ -95,11 +105,20 @@ class Dashboard extends Component {
   }
 
   render() {
-    const { pluginList, loading } = this.state;
+    const { pluginList, loading, error } = this.state;
     const { store } = this.props;
     const userName = store.get('userName') || '';
     return (
       <React.Fragment>
+        {error && (
+          <Notification
+            title={error} 
+            position='top-right' 
+            variant='danger' 
+            closeable
+            onClose={()=>this.setState({error:null})} 
+          />
+        )}
         <div className="plugins-stats">
           <div className="row plugins-stats-row">
             <div className="title-bar">{`Dashboard for ${userName}`}</div>
--- a/src/components/Dashboard/Dashboard.js
+++ b/src/components/Dashboard/Dashboard.js
@@ -70,6 +70,9 @@ class Dashboard extends Component {
       if(response.data) {
         this.fetchPlugins();
       }
+      else{
+        throw new Error('Delete unsuccessful');
+      }
     } catch (e) {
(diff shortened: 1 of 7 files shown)


src/errors/HttpApiCallError.js
```diff
@@ -0,0 +1,9 @@
+export default class HttpApiCallError extends Error {
+  constructor(error) {
+    super(error);
+    this.message = error.message;
```
jennydaman, line 4: Could use object spread syntax here (optional).

src/components/Plugins/Plugins.js
```diff
@@ -52,7 +55,8 @@ export class Plugins extends Component {
 
   componentDidMount() {
     this.fetchPlugins().catch((err) => {
-      console.error(err);
+      console.log("here")
```
jennydaman, line 58: delete `console.log`

jennydaman (commented):
Only nitpicks, see review. 

Preview

src/components/Dashboard/Dashboard.js
```diff
@@ -61,6 +69,7 @@ class Dashboard extends Component {
         this.fetchPlugins();
       });
     } catch (e) {
+      this.showNotifications(new HttpApiCallError(e));
       return e;
```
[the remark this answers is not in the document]
DivyanshiSingh: We can either use custom messages as I was using earlier or use the custom response from the backend, but as you said that the backend response is not structured, I think it's best to use the custom messages.

src/components/Plugins/Plugins.js
DivyanshiSingh: console deleted.

src/components/Dashboard/Dashboard.js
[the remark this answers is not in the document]
jennydaman: We can detect some custom messages simply based on the response (404, wrong content type, timeout, ...)
However that is besides the point. What you added is okay. I was making a comment on the line below, how the code reads

```javascript
catch (e) {
  something();
  return e;
}
```

In general, we shouldn't return anything in the case of an exception, we should either throw it again or give up and do nothing. If we return an exception, we are making a bad situation worse.

However, I took a moment to scroll up and actually read the existing code. It's async spaghetti. The `try` is only wrapping these lines of code:

https://github.com/FNNDSC/ChRIS_store_ui/pull/144/files#diff-e82a9b05e2166cb4373028e8db04cb652be90796eb324ad21dbbf2f5ef9588a9L30-L46

The actual request and errors regarding the request occur during a `Promise`.

I think what was intended here was,

```javascript
client.getPlugin(pluginId)
  .then(plugin => plugin.delete())
  .then(() => this.fetchPlugins())
  .catch(e => this.showNotifications(new HttpApiCallError(e)));
```

Can be cleaner using `async`/`await`. Do you think you could try rewriting this using `async`/`await`?

jennydaman (changes requested):

src/components/Dashboard/Dashboard.js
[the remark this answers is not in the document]
DivyanshiSingh: thanks for pointing this out. I will remove the return statement and rewrite using async/await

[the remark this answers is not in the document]
DivyanshiSingh: changes done please review

src/components/Dashboard/Dashboard.js
```diff
@@ -49,20 +57,21 @@ class Dashboard extends Component {
...
     const storeURL = process.env.REACT_APP_STORE_URL;
     const auth = { token: store.get('authToken') };
     const client = new Client(storeURL, auth);
 
     let response;
     try {
-      response = client.getPlugin(pluginId).then(plugin => plugin.delete());
-      response.then(() => {
+      response = await client.getPlugin(pluginId);
+      response.delete();
```
jennydaman, line 69: Let's  `await response.delete()` as well.

jennydaman (changes requested):
Sorry for the late responses @DivyanshiSingh. This PR is almost ready.

src/components/Dashboard/Dashboard.js
manas2297: done

src/components/Dashboard/Dashboard.js
```diff
@@ -49,20 +57,21 @@ class Dashboard extends Component {
...
     let response;
     try {
-      response = client.getPlugin(pluginId).then(plugin => plugin.delete());
-      response.then(() => {
+      response = await client.getPlugin(pluginId);
+      await response.delete();
+      if(response.data) {
         this.fetchPlugins();
-      });
+      }
```
jennydaman, line 72: else, maybe throw an error to trigger UI notification?

src/components/Plugin/Plugin.js
```diff
@@ -41,12 +44,15 @@ export class Plugin extends Component {
...
     let pluginData;
-    if (!this.state.pluginData) {
-      pluginData = await this.fetchPluginData();
-    } else {
-      ({ pluginData } = this.state);
+    try{
+      if (!this.state.pluginData) {
+        pluginData = await this.fetchPluginData();
+      } else {
+        ({ pluginData } = this.state);
```
jennydaman, line 51: Would you remove these parenthesis?

src/components/Plugin/Plugin.js
```diff
@@ -114,18 +126,20 @@ export class Plugin extends Component {
...
+    try{
+      if (auth) {
+        const client = new Client(storeURL, auth);
+        const response = await client.getPluginStars({ plugin_name: pluginName });
+        stars = response.data;
+      }
+  
+      if (stars.length > 0) {
+        this.setState({ star: stars[0] });
+      }
```
jennydaman, line 138: else... throw

DivyanshiSingh: Can you specify more about what to throw here?

DivyanshiSingh: done

src/components/Plugin/Plugin.js
DivyanshiSingh: done

src/components/Dashboard/Dashboard.js
DivyanshiSingh: done

jennydaman (approved):
I myself would be more specific about errors and exception handling (either "unexpected response from server," or in the worst case, "unknown error"), but being more specific about error handling, though it might not be helpful at this level. What you have here greatly improves UX and is good to merge.

DivyanshiSingh:
Hey, @jennydaman do you have some specific feature in mind that you feel is primarily important to add in pf4. I can take that up in parallel to form. I will raise the PR for the form by today.

jennydaman:
@DivyanshiSingh it's definitely still about eliminating all trace of `patternfly-react` from the codebase, and maybe upgrading to react v17. They're chores but we must pay off the technical debt before working on the features.

PR #147, opened 2021-04-14
Functionality to sort searches

sonylomo:
Added functionality to sort search:

-   alphabetically by author
-   alphabetically by name
-   chronologically by date created

### DIFF
--- a/src/components/Plugins/Plugins.js
+++ b/src/components/Plugins/Plugins.js
@@ -1,15 +1,14 @@
-import React, { Component } from 'react';
-import PropTypes from 'prop-types';
-import Client from '@fnndsc/chrisstoreapi';
-import { DropdownButton, MenuItem } from 'react-bootstrap';
-import PluginItem from './components/PluginItem/PluginItem';
-import LoadingPluginItem from './components/LoadingPluginItem/LoadingPluginItem';
-import PluginsCategories from './components/PluginsCategories/PluginsCategories';
-import './Plugins.css';
-import LoadingContainer from '../LoadingContainer/LoadingContainer';
-import LoadingContent from '../LoadingContainer/components/LoadingContent/LoadingContent';
-import ChrisStore from '../../store/ChrisStore';
-
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import Client from "@fnndsc/chrisstoreapi";
+import { DropdownButton, MenuItem } from "react-bootstrap";
+import PluginItem from "./components/PluginItem/PluginItem";
+import LoadingPluginItem from "./components/LoadingPluginItem/LoadingPluginItem";
+import PluginsCategories from "./components/PluginsCategories/PluginsCategories";
+import "./Plugins.css";
+import LoadingContainer from "../LoadingContainer/LoadingContainer";
+import LoadingContent from "../LoadingContainer/components/LoadingContent/LoadingContent";
+import ChrisStore from "../../store/ChrisStore";
 
 // ==============================
 // ------ PLUGINS COMPONENT -----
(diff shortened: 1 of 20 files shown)


src/components/Plugins/Plugins.js
```diff
@@ -196,7 +201,21 @@ export class Plugins extends Component {
...
     if (pluginList) {
-      pluginListBody = pluginList.map((plugin) => (
+      const sorted = pluginList.sort((a, b) => {
+        if (sortType === "name") {
+          return a[sortType] > b[sortType] ? 1 : -1;
+        }
+        if (sortType === "authors") {
+          return a[sortType] > b[sortType] ? 1 : -1;
+        }
+        if (sortType === "creation_date") {
```
jennydaman, line 211: How come two different approaches to comparing `creation_date` are being used here?

Is `creation_date` a string? If so, maybe better to create a [`Date`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) object and compare those.

src/components/Plugins/Plugins.js
```diff
@@ -196,7 +201,21 @@ export class Plugins extends Component {
 
     // Render the pluginList if the plugins have been fetched
     if (pluginList) {
-      pluginListBody = pluginList.map((plugin) => (
+      const sorted = pluginList.sort((a, b) => {
```
jennydaman, line 204: A subjectively cleaner implementation might store the comparer function in the state. e.g., say I want to sort by name. After clicking "Name" the code is called

```javascript
this.setState({sortFunc: (a, b) => a.name > b.name ? 1 : -1});
```

Here you would simply call:

```javascript
const sorted = pluginList.sort(sortFunc);
```

Saves you from an if/then ladder and having types manage themselves.

src/components/Plugins/Plugins.js
sonylomo: The last comparison of `creation_date` is to return the default arrangement of plugins: from the latest created to earliest.

src/components/Plugins/Plugins.js
sonylomo: I'll try that out

src/components/Plugins/Plugins.js
sonylomo: The cleaner format you've suggested sorts this out.

src/components/Plugins/Plugins.js
sonylomo: I've implemented your suggestion but the plugins component re-renders twice and it's causing some tests to fail.

According to [this article](https://mariosfakiolas.com/blog/my-react-components-render-twice-and-drive-me-crazy/), it's a `REACT.StrictMode` thing to help out with debugging in development but it won't occur in production. Should I change the tests to accommodate this?

jennydaman: That article was very helpful. How do you plan on changing the tests?

sonylomo: failed test 1
failed test 2

I thought that instead of : 
`it('should render the title prop for each Plugin', () => {
    plugins.forEach((plugin, i) => {
      expect(plugin.props.title).toEqual(`testTitle${i + 1}`);
    });
  });`

it could equal to `testTitle${i + 2}` or just have `testTitle${ i }`. But never mind, it doesn't work 🙃.

sonylomo: And now I'm stuck because of tests...again 😅.

jennydaman: Do tests happen against the production or development version of a build? I've never considered that question before...

If you're super stuck, push what you have so I can take a look.

Issue #135, opened 2021-04-11
Sorting of plugin search results


jennydaman:
https://github.com/FNNDSC/ChRIS_store_ui/blob/ca45de40784ddb979445252c3d71c05bc3699b50/src/components/Plugins/Plugins.js#L218

Some sorts could be:

- alphabetical, by author
- alphabetical, by name
- chronological, by date created

sonylomo:
Can I work on this one @jennydaman?

sonylomo:
![conflict](https://user-images.githubusercontent.com/49971500/114519936-928a1280-9c49-11eb-88df-e54337479c6c.png)
I'm trying to resolve conflicts but my code isn't passing these 2 tests. I didn't change any code related to them. I've checking what's wrong and updated my branch to be inline with the master before fixing plugin sort.

I've also checked if the  "mark as favorite" functionality on plugins store website is working but it's not - was it working before?

jennydaman:
Try this:

```bash
git remote add upstream git@github.com:FNNDSC/ChRIS_store_ui.git
git fetch upstream
git diff upstream/master
```

I found this:

```patch
diff --git a/src/components/Plugins/Plugins.js b/src/components/Plugins/Plugins.js
index a157f49..e5e2b45 100755
--- a/src/components/Plugins/Plugins.js
+++ b/src/components/Plugins/Plugins.js
@@ -113,20 +114,23 @@ export class Plugins extends Component {
     }
   }

-  fetchPlugins = () => {
-    const params = new URLSearchParams(window.location.search)
-    const name = params.get('q') //get value searched from the URL
+  fetchPlugins() {
+    const params = new URLSearchParams(window.location.search);
+    const name = params.get("q"); //get value searched from the URL
+
     const searchParams = {
       limit: 20,
       offset: 0,
       name_title_category:name,
-    };
+    }
+
```

Notice the declaration of `fetchPlugins` was changed. This is an oddity with syntax. If a function is defined as `fetchPlugins() {`, `this` will be undefined unless you manually "bind" it.

sonylomo:
> Notice the declaration of `fetchPlugins` was changed. This is an oddity with syntax. If a function is defined as `fetchPlugins() {`, `this` will be undefined unless you manually "bind" it.

Thanks for that. What about the `starsByPlugin` test?

jennydaman:
[...]

PR #153, opened 2021-04-17
Sort fix

sonylomo:
Fixes issue #135 
Added functionality to sort search:

-   alphabetically by author
-   alphabetically by name
-   chronologically by date created

### DIFF
--- a/src/components/Plugins/Plugins.js
+++ b/src/components/Plugins/Plugins.js
@@ -1,15 +1,14 @@
-import React, { Component } from 'react';
-import PropTypes from 'prop-types';
-import Client from '@fnndsc/chrisstoreapi';
-import { DropdownButton, MenuItem } from 'react-bootstrap';
-import PluginItem from './components/PluginItem/PluginItem';
-import LoadingPluginItem from './components/LoadingPluginItem/LoadingPluginItem';
-import PluginsCategories from './components/PluginsCategories/PluginsCategories';
-import './Plugins.css';
-import LoadingContainer from '../LoadingContainer/LoadingContainer';
-import LoadingContent from '../LoadingContainer/components/LoadingContent/LoadingContent';
-import ChrisStore from '../../store/ChrisStore';
-
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import Client from "@fnndsc/chrisstoreapi";
+import { DropdownButton, MenuItem } from "react-bootstrap";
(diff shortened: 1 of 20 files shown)


src/components/Plugins/Plugins.js
```diff
@@ -24,6 +24,7 @@ export class Plugins extends Component {
 
     this.mounted = false;
     this.state = {
+      sortFunc: (a, b) => new Date(a.creation_date) + new Date(b.creation_date),
```
jennydaman, line 27: Did you mean to use the `-` operand?

src/components/Plugins/Plugins.js
```diff
@@ -42,20 +43,18 @@ export class Plugins extends Component {
...
       /**
        * Accumulate counts of categories from fetched plugins
        */
       let { categories } = this.state;
       for (const [index, { name, length }] of categories.entries()) {
         categories[index].length = plugins.reduce(
-          (count, current) => (name === current.category) 
-            ? ++count
-            : count, 
+          (count, current) => (name === current.category ? ++count : count),
```
jennydaman, line 53: A more efficient algorithm can be used here.

Currently you have O(n^2):

```
for every category:
  for every plugin:
    if plugin's category is ..., then increase counter
```

It could be linear, using a `Map` with category name as keys and frequency as values.

```
for every plugin:
  increase count of plugin's category
```

jennydaman (changes requested):

src/components/Plugins/Plugins.js
sonylomo: No. I wanted it to be sorted from the latest creation date to the earliest by default. The other way would've been `sortFunc: undefined` - I didn't think that would be a good idea 😅.

jennydaman: You have a positive number plus a positive number will always result in a positive number, so this function isn't doing anything.

Maybe you meant

```javascript
// Monday, Tuesday
sortFunc: (a, b) => new Date(a.creation_date) - new Date(b.creation_date)
// Tuesday, Monday
sortFunc: (a, b) => new Date(b.creation_date) - new Date(a.creation_date)
```

sonylomo:
The tests on `LoadingContainers` are failing, I'm not sure why. They were previously passing.

jennydaman:
Closing because `pf4` was merged into `master`

Issue #160, opened 2021-04-18
App crashes when pluginId is not valid


zrthxn:
## Description
The store app crashes when the pluginId in the URL is not valid. This can happen when a `Link` is used instead of an `a` anchor tag which puts a URL in the path which is not a valid ID.

![Screenshot (53)](https://user-images.githubusercontent.com/35369637/115143504-60234f80-a065-11eb-8335-86aee95ae9f6.png)

## Cause
The function `fetchPluginData()` in `src/components/Plugin/Plugin.js` is called by the `componentDidMount` lifecycle method. 
https://github.com/FNNDSC/ChRIS_store_ui/blob/f5fbc1b65f8a9c8e6d538a4ce0f7f8b7348949d8/src/components/Plugin/Plugin.js#L42-L53

However, this function can throw an unhandled Promise rejection if the ID is invalid.
https://github.com/FNNDSC/ChRIS_store_ui/blob/f5fbc1b65f8a9c8e6d538a4ce0f7f8b7348949d8/src/components/Plugin/Plugin.js#L98-L104

PR #161, opened 2021-04-18
Updated plugin body

zrthxn:
## Overview of changes
These fixes focus on making the Plugin component more relevant and useable.
1. Fetching repository data from Github public API to use for populating relevant fields.
2. Fetch README from public repo community profile from Github public API to display as plugin documentation.
3. Display README as Markdown using package [marked](https://www.npmjs.com/package/marked).
4. Fixed many layout issues in PluginBody and made the component slightly more mobile friendly.

### Images
Screenshot (52)

## Work in Progress
With further discussion more changes will be added according to what direction is to be taken. 
This is not the final state of this work.

### DIFF
--- a/src/components/Plugin/Plugin.js
+++ b/src/components/Plugin/Plugin.js
@@ -183,14 +183,14 @@ export class Plugin extends Component {
             <Grid>
               <Grid.Row>
                 <Grid.Col sm={12}>
-                  <Grid.Col sm={1}>
+                  <Grid.Col sm={2}>
                     <img
                       className="plugin-icon"
                       src={PluginImg}
                       alt="Plugin icon"
                     />
                   </Grid.Col>
-                  <Grid.Col sm={6}>
+                  <Grid.Col sm={10}>
                     <div className="plugin-category">
                       Visualization
                     </div>
@@ -207,24 +207,24 @@ export class Plugin extends Component {
                     <div className="plugin-description">
                       {data.description}
                     </div>
-                  </Grid.Col>
-                  <Grid.Col sm={4} className="plugin-stats">
-                    <Icon name="star" size="lg" />
-                    {' '}
-                    10k+
-                    {modificationDate.isValid()
-                      && (
+                    <div className="plugin-stats">
+                      <Icon name="star" size="lg" />
+                      {' '}
+                      10k+
+                      {modificationDate.isValid()
+                        && (
+                        <span className="plugin-modified">
+                          <Icon name="clock-o" size="lg" />
+                          {`Last modified ${modificationDate.format()}`}
+                        </span>
+                        )}
+                      {/* temp text */}
                       <span className="plugin-modified">
                         <Icon name="clock-o" size="lg" />
-                        {`Last modified ${modificationDate.format()}`}
+                        {' '}
+                        Last modified 1 hour ago
                       </span>
-                      )}
-                    {/* temp text */}
-                    <span className="plugin-modified">
-                      <Icon name="clock-o" size="lg" />
-                      {' '}
-                      Last modified 1 hour ago
-                    </span>
+                    </div>
                   </Grid.Col>
                 </Grid.Col>
               </Grid.Row>
--- a/src/components/Plugin/Plugin.js
+++ b/src/components/Plugin/Plugin.js
@@ -1,18 +1,20 @@
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
-import { Grid, Icon } from 'patternfly-react';
+import { Grid, GridItem } from '@patternfly/react-core';
+import { StarIcon, ClockIcon } from '@patternfly/react-icons';
 import PropTypes from 'prop-types';
 import Client from '@fnndsc/chrisstoreapi';
+
 import LoadingPlugin from './components/LoadingPlugin/LoadingPlugin';
 import PluginBody from './components/PluginBody/PluginBody';
 import RelativeDate from '../RelativeDate/RelativeDate';
 import ChrisStore from '../../store/ChrisStore';
 import PluginImg from '../../assets/img/brainy-pointer.png';
 import NotFound from '../NotFound/NotFound';
-
-import './Plugin.css';
 import Notification from '../Notification';
 import HttpApiCallError from '../../errors/HttpApiCallError';
+
+import './Plugin.css';
 
 const removeEmail = (author) => {
   if (author) return author.replace(/( ?\(.*\))/g, '');
@@ -164,7 +166,7 @@ export class Plugin extends Component {
       className = 'plugin-star-disabled';
       name = 'star-o';
     }
-    return <Icon name={name} className={className} onClick={this.onStarClicked} />;
+    return <StarIcon name={name} className={className} onClick={this.onStarClicked} />;
   }
 
   render() {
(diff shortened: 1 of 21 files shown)


zrthxn:
Fixes: #160 

Handle any Promise rejection by `fetchPluginData()` in the `componentDidMount` lifecycle method of `src/components/Plugin/Plugin.js`. Return a 404 screen if after loading there is no pluginData.

Screenshot (54)

jennydaman:
Awesome. I'll review when you say it's ready to go.

We've had several discussions about how best to address the README issue, this approach has a good effort-to-reward solution. We can think about handling other issues (like READMEs for outdated plugins) in the future.

zrthxn:
There's a slight problem though, that most existing plugins have readme files in .rst format which doesn't show up perfectly as Markdown.

zrthxn:
Added temporary support for .rst files but the rendered output doesn't look great. So currently treating it as Markdown.

zrthxn:
@jennydaman You can review this now

jennydaman:
@zrthxn RST is out of scope, don't worry about it

src/components/Plugin/Plugin.js
```diff
@@ -90,24 +98,22 @@ export class Plugin extends Component {
...
-        pluginData.url = plugin.url;
-      } catch (e) {
-        return reject(e);
-      }
+    try {
+      const plugin = await this.client.getPlugin(parseInt(pluginId, 10));
+      pluginData = plugin.data;
+      pluginData.url = plugin.url;
+    } catch (e) {
+      return Promise.reject(e);
```
jennydaman, line 110: I think you should just not bother catching here if the exception cannot be handled by this function.

I am not sure it makes sense to `return Promise.reject(e)`, especially not in an `async` block

src/components/Plugin/components/PluginBody/PluginBody.js
```diff
@@ -1,209 +1,167 @@
...
-                              <a className="pf-c-button pf-m-link">
-                                View all contributors...
-                              </a>
+const PluginBody = ({ pluginData }) => {
+  const [readme, setReadme] = useState();
+  const [repoData, setRepoData] = useState();
+
+  const getReadmeFileType = (url) => (
+    url
+      .split('').reverse().join('')   // Reverse the URL
```
jennydaman, line 20: This is very inefficient.

Since README is being fetched by URL, could you try checking its `Content-Type`?

jennydaman (changes requested):

src/components/Plugin/Plugin.js
zrthxn: Returning a rejected Promise here is so that the catch error can be "bubbled up" to the componentDidMount where this function is called and then it can be handled by the catch block there. This way all errors can be handled in the same place (i.e. the catch block in the CDM method)

https://github.com/FNNDSC/ChRIS_store_ui/blob/63d9b5d84e1620048721abe54d059c42b41b2f02/src/components/Plugin/Plugin.js#L48-L52

If this seems like an anti-pattern then the errors can be handled where they happen.

src/components/Plugin/components/PluginBody/PluginBody.js
zrthxn: Unfortunately the downloads from raw.githubusercontent.com (which is where the `download_url` points to) do not supply the appropriate Content Type header for README files. Its always `text/plain; charset=utf-8`.

src/components/Plugin/Plugin.js
jennydaman: `return` _typically_ implies that the state is expected and that the data is sane. The safer way of handling errors would be to let exceptions be thrown and caught. To handle this exception, wouldn't you need something like

```javascript
let result = await fun();
if (result instanceof Promise.reject) ...
```

jennydaman: In other words, exceptions mean "something unexpected happened, and the app's state (data) doesn't make sense anymore." Thus when a function does not throw an exception, it means "everything worked as expected." If you catch an exception without being able to restore the state (in this case, set `loading: false` and get stars) then the component won't know that something went wrong, and it'll keep trying to work on bad data (an unexpected `Promise.reject` object) leading to more data/state corruption.

Catch an exception if:

- the state can be corrected somehow (e.g. ok, something went wrong, let's show an error message and stop making things worse)
- within the function's scope it can handle the exception in a way that is better than how its parent might handle the exception
- we can enhance the exception before rethrowing it with extra information from the current scope

src/components/Plugin/components/PluginBody/PluginBody.js
```diff
@@ -1,209 +1,148 @@
...
-                              {/* eslint-disable-next-line */}
-                              <a className="pf-c-button pf-m-link">
-                                View all contributors...
-                              </a>
+const PluginBody = ({ pluginData }) => {
+  const [readme, setReadme] = useState();
+  const [repoData, setRepoData] = useState();
+
+  const fetchReadme = useCallback(async (repo) => {
+    const url = (await
```
jennydaman, line 16: Please flatten this...

Or, you might be interested in "pipe syntax." Currently it is just a [proposal to the language](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Pipeline_operator), but there are NPM packages implementing the functionality (look them up) --- and you could implement it yourself in 10-20 lines of code.

src/components/Plugin/components/PluginBody/PluginBody.js
```diff
@@ -1,209 +1,148 @@
...
+        await (
+          await fetch(`https://api.github.com/repos/${repo}/community/profile`)
+        ).json()).files.readme.url)
+      ).json()).download_url
+
+    const file = await fetch(url)
+    const type = url.split('.').reverse().shift()
+
+    if (type === 'md' || type === 'rst') 
+      setReadme(marked(await file.text()))
```
jennydaman, line 27: I would recommend moving `setReadme` outside of `fetchReadme`.

Doing so adheres to "functional programming" and has a few advantages:

- the function `fetchReadme` becomes "stateless," i.e. without side effects — fewer bugs
- a stateless function which processes data deterministically is easier to test (you can write unit tests to test the functionality of fetching a README without having to test how it renders the README into JSX)
- separate data processing logic (the "model") from logic related to updating the UI (the "controller") see [MVC pattern](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller). Separation makes it easier to replace one part (say, replacing the visual layer without changing how the data processing logic works, as we're trying to do right now with the Patternfly 3 -> 4 upgrade) easier.

src/components/Plugin/components/PluginBody/PluginBody.js
```diff
@@ -1,209 +1,148 @@
...
+      fetchReadme(repo)
+    } catch (error) {
+      console.error(error)
+      throw Error(error)
+    }
+  }, [fetchReadme])
+
+  useEffect(() => {
+    const expectRepoName = pluginData.public_repo.split('github.com/')[1];
+    if (expectRepoName)
```
jennydaman, line 45: Did you `yarn run lint`?

src/components/Plugin/components/PluginBody/PluginBody.js
```diff
@@ -1,209 +1,148 @@
...
+      try {
+        fetchRepoData(expectRepoName)
+      } catch (error) {
+        console.error(error)
```
jennydaman, line 49: https://github.com/FNNDSC/ChRIS_store_ui/pull/161#discussion_r618080232

src/components/Plugin/components/PluginBody/PluginBody.js
```diff
@@ -1,209 +1,148 @@
...
+                        </NavItem>
+                      </Nav>
+                      <TabContent animation className="plugin-tab">
+                        <TabPane eventKey={1}>
+                          <div id="plugin-body-container">
+                            <div className="plugin-body-main-col">
+                              <div className="plugin-body-readme">
+                                README
+                              </div>
+                              { readme ? <div dangerouslySetInnerHTML={{ __html: readme }}></div> : null }
```
jennydaman, line 85: Please see the security warning about the `marked` package.
https://www.npmjs.com/package/marked

jennydaman (changes requested):

src/components/Plugin/components/PluginBody/PluginBody.js
zrthxn: Yes and it didn't give me any problems related to this?
Screenshot (56)

src/components/Plugin/components/PluginBody/PluginBody.js
zrthxn: Added sanitization on the HTML generated by marked. Although I'm not completely sure that this would've introduced an XSS vulnerability since we are getting the content from Github and I think it does sanitize the README. Not sure though.

src/components/Plugin/components/PluginBody/PluginBody.js
zrthxn: Moved `setReadme` outside of `fetchReadme`
Moved `setRepoData` outside of `fetchRepoData`

src/components/Plugin/Plugin.js
zrthxn: Alright I understand. It was wrong to throw the same Exception twice. I've now switched to handling the errors where they happen, if they can be, otherwise let the function reject the Promise.

src/components/Plugin/components/PluginBody/PluginBody.js
zrthxn: I didn't know about that, I'll look into it. Although in this case it was simpler to just refactor a bit.

src/components/Plugin/components/PluginBody/PluginBody.js
jennydaman: I must be getting eslint configs confused across different projects, sorry...

src/components/Plugin/components/PluginBody/PluginBody.js
jennydaman: Github would sanitize markdown when they render it themselves, but they serve it as-is when you request the raw `.md` file.

src/components/Plugin/components/PluginBody/PluginBody.js
```diff
@@ -1,209 +1,167 @@
...
+
+  useEffect(() => {
+    async function fetchRepo() {
+      try {
+        const data = await fetchRepoData(expectRepoName)
+        setRepoData(data)
+
+        const { file, type } = await fetchReadme(expectRepoName)
+        if (type === 'md' || type === 'rst') 
+          setReadme(marked(file))
```
jennydaman, line 47: Security paranoia: I would call `sanitize` somewhere around here. That way, the untrusted data is never stored in state. (Also marginally more efficient in situation where state can be set preemptively, i.e. before `render()`.)

zrthxn: That does make sense, also prevents accidental usage of unsafe html.

src/components/Plugin/components/PluginBody/PluginBody.js
```diff
@@ -1,209 +1,168 @@
 /* eslint-disable react/jsx-indent-props */
 /* eslint-disable react/jsx-indent */
-import React from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
+import { Grid, Nav, NavItem, TabContainer, TabContent, TabPane } from 'patternfly-react';
```
jennydaman, line 4: Would you try converting this page to use components from `@patternfly/react-core`?

jennydaman (changes requested):

src/components/Plugin/components/PluginBody/PluginBody.js
zrthxn: Done

jennydaman (approved):
I still think [`src/components/Plugin/components/PluginBody/PluginBody.js`](https://github.com/FNNDSC/ChRIS_store_ui/pull/161/files#diff-198b4e0e2d3645575d568b23f6cad63b62fdd14d3c3c6e790059c8ec1ee3915f) should be split up to conform with a MVC pattern. Nonetheless, this feature is very important and I think you did a good job with what's here.

Do you want it to be merged now and leave refactoring as subsequent work?

zrthxn:
For the moment I think it's working well enough to be tested. But if you'd like to get it completed then I don't mind. About MVC, I think that might be overworking something that doesn't need that level of complication?

jennydaman:
Good point. I suppose getting READMEs from Github is a solution "for the time being" anyways, since eventually we'll need to implement a better solution which takes into account outdated plugins and plugins hosted on other websites.

Let me do some more testing and then I will be happy to merge.

zrthxn:
A few things I think could be done to support other Git platforms/providers and handle versions. 
1. Support a limited set of git platforms, the most popular ones like Github, gitlab, bitbucket etc. Since it would be realistically impossible to support every single one. 
2. Instead of using the git repository, use DockerHub to manage readme/documentation since it's a central platform and any image that can be pulled by tag would have to be there. 
3. Resolve the issues with plugins getting outdated after discussing a bit more.

PR #171, opened 2021-04-25
Fix search qs and optimize Plugins.js

jennydaman:
@DivyanshiSingh and @zrthxn would you review these changes?

@DivyanshiSingh I fixed the search functionality to correctly set the query string.

@zrthxn I did a minor overhaul of `Plugins.js`, documenting it (to my understanding) and some optimizations.

- minimized calls to `setState`: it is called only once after fetching plugin data, counting category frequencies, and then getting stars.
- name of selected category is used as a filter instead of mutating `pluginList`
  - minimizes API calls
- de-spaghetti-fy async/await

### DIFF
--- a/src/components/Navbar/components/Search/Search.js
+++ b/src/components/Navbar/components/Search/Search.js
@@ -75,11 +75,7 @@ const Search = (props) => {
               <li
                 key={id}
                 data-id={item.id}
-                onMouseDown={(e) => {
-                  history.push({
-                  pathname: '/plugins',
-                  search: item.name,
-                })}}
+                onMouseDown={(e) => {onSearch(value, 'ENTER')}}
               >
                 {item.name}
               </li>
--- a/src/components/Navbar/components/Search/Search.js
+++ b/src/components/Navbar/components/Search/Search.js
@@ -75,7 +75,7 @@ const Search = (props) => {
               <li
                 key={id}
                 data-id={item.id}
-                onMouseDown={(e) => {onSearch(value, 'ENTER')}}
+                onMouseDown={(e) => {onSearch(item.name, 'ENTER')}}
               >
                 {item.name}
               </li>
--- a/src/components/Plugins/Plugins.js
+++ b/src/components/Plugins/Plugins.js
@@ -14,11 +14,12 @@ import Notification from '../Notification';
 
 import './Plugins.css';
 
+const CATEGORIES = ['FreeSurfer', 'MRI', 'Segmentation', 'copy'];
 
-// ==============================
-// ------ PLUGINS COMPONENT -----
-// ==============================
-
+/**
+ * A page showing a list of ChRIS plugins, according to the search
+ * specified in the URI's query string.
+ */
 export class Plugins extends Component {
   constructor(props) {
     super(props);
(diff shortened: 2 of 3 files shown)


DivyanshiSingh:
Yeah sure

jennydaman:
There is also a lot of room for optimizing the favorite plugins data structure.

src/components/Plugins/Plugins.js
```diff
@@ -27,61 +28,41 @@ export class Plugins extends Component {
...
-      });
+
+  /**
+   * Re-fetch the list of plugins if the input was changed
+   * in the NarBar's search bar.
+   */
+  async componentDidUpdate(prevProps) {
+    if (this.props.location !== prevProps.location) {
+      await this.refreshPluginList();
     }
```
zrthxn, line 56: I don't think these two lifecycle methods (componentDidMount and componentDidUpdate) need to be async just to await a call to an async function, since there is nothing happening in those functions after the call to the `refreshPluginList` function. In other words if they dont depend on the return values from `refreshPluginList` they dont need to "await" it.

src/components/Plugins/Plugins.js
```diff
@@ -91,16 +72,33 @@ export class Plugins extends Component {
...
+  /**
+   * Show a notification for some network error.
+   * @param error error message
+   */
   showNotifications = (error) => {
-    console.log(error.message)
+    console.error(error);
     this.setState({
       errors: error.message,
     })
```
zrthxn, line 92: This might just be semantics but I thought that if the state variable is called "errors" then it should be an array of errors and this function should push to that array. This will allow multiple notifications to be shown.

src/components/Plugins/Plugins.js
```diff
@@ -131,72 +135,85 @@ export class Plugins extends Component {
...
-      this.showNotifications(new HttpApiCallError(error));
+    // count the frequency of plugins which belong to categories
+    for (const { category } of plugins.data) {
+      // TODO make category counting case insensitive
+      console.log(category);
+      const currentCount = categories.get(category);
+      if (currentCount !== undefined) {
+        categories.set(category, currentCount + 1);
+      }
     }
```
zrthxn, line 176: This can be refactored to the following.
```js
for (const { category } of plugins.data) {
    // TODO make category counting case insensitive
    console.log(category);
    categories.set(category, categories.has(category) ? categories.get(category) + 1 : 0);
}
```
This will not only account for categories which are not in the Map (i.e. `currentCount !== undefined`) but also for null values, empty strings etc.

zrthxn (changes requested):

src/components/Plugins/Plugins.js
jennydaman: I also gotta take out the `console.log`

My implementation was an intentional design choice but now I am double-thinking: should a fixed list of categories be shown, or should all discovered categories be shown? I will probably ask @rudolphpienaar and @mairin for their input.

The reason I've avoided `categories.has` is just to achieve one fewer `Map` invocation. (I can't justify myself for treating web JS as embedded systems).

src/components/Plugins/Plugins.js
jennydaman: It's a good call, variable names should make sense.
@DivyanshiSingh are multiple notifications supported? Either way, for simplicity here I will change it to only use one error notification.

zrthxn: I dont think they are but the way I've handled it [here](https://github.com/zrthxn/ChRIS_store_ui/blob/8e51bd22e44c61978e7beed3095f516f5dc12140/src/components/Plugin/Plugin.js#L260-L272) seems to work.

src/components/Plugins/Plugins.js
jennydaman: You're correct.

An argument can be made both ways to justify either practice. By leaving out `await`, we have more minimal code which means it's better understood and perhaps cleaner.

I used `await` because my IDE told me to do so. There is a reason for this — asynchronous programming can be understood at two levels: "fire and forget" concurrency, or "Promises." Of course the latter is more accurate to JavaScript's model, since `Promise` is the implementation of `async`. Conceptually, a `Promise` is the expectation for something to happen (data synthesis or a state change) in the future. All `Promise` (besides one top-level `Promise`) should be handled (by `.then()` or `await`). An unhandled `Promise` becomes orphaned: it will produce data (or a state change) that no code can respond to. It might be the case that `React` internals would want to `await` on the lifecycle methods. I don't think it actually does, but nonetheless we should leave the decision to orphan the `Promise` to React's internals instead of orphaning it in our code.

src/components/Plugins/Plugins.js
jennydaman: In what situation might multiple error messages appear around the same time?

In your solution, do the notifications overlap or do they show correctly?

DivyanshiSingh: > It's a good call, variable names should make sense.
> @DivyanshiSingh are multiple notifications supported? Either way, for simplicity here I will change it to only use one error notification.

It's not supported for multiple notifications. At the same time, the solution by @zrthxn will overlap. If you want I can make the changes with the correct implementation.

jennydaman: No need. I don't think there will be a situation where we need to show multiple error notifications.

zrthxn: > In what situation might multiple error messages appear around the same time?
> 
> In your solution, do the notifications overlap or do they show correctly?

In my case they get staggered by the timing of the requests but without that they would overlap on top of each other.

src/components/Navbar/components/Search/Search.js
```diff
@@ -75,11 +75,7 @@ const Search = (props) => {
               <li
                 key={id}
                 data-id={item.id}
-                onMouseDown={(e) => {
-                  history.push({
-                  pathname: '/plugins',
-                  search: item.name,
-                })}}
+                onMouseDown={(e) => {onSearch(value, 'ENTER')}}
```
DivyanshiSingh, line 78: Here, onMouseDown prop is to select a particular item. So, the parameter passed would be onSearch(item.name, 'ENTER')

DivyanshiSingh (changes requested):

src/components/Navbar/components/Search/Search.js
jennydaman: Wow I pushed the commit at the same time you made this comment!

DivyanshiSingh: did you make the changes ? or else it will behave in a wrong way

jennydaman: https://github.com/FNNDSC/ChRIS_store_ui/pull/171/commits/ff17bd9d78d4e2dbb2efcbe8fac169f50c76bcc6

Issue #166, opened 2021-04-24
Application crash on going to a particular plugin 


DivyanshiSingh:
On routing to a particular plugins page, if the 'fetchRepoData' function fails the application is crashing.
![image](https://user-images.githubusercontent.com/53593872/115964968-90c32780-a544-11eb-9a19-4c418ccad9e1.png)

This error happened because the API (https://api.github.com/repos/${repo}) fails.
@jennydaman please check this, I can work on it.

jennydaman:
I will take a look at it.

Related to #161 so tagging @zrthxn

jennydaman:
@DivyanshiSingh can you provide more specific instructions on how to reproduce?

DivyanshiSingh:
Route to /plugins then choose any plugin listed, as soon as the plugins detail page loads, the application crashes

jennydaman:
Which branch, master or pf4? Have you tried `rm -rf node_modules && yarn`?

zrthxn:
Shouldn't happen on pf4 since this was something I knew about and I had fixed. I'll check it out again.

DivyanshiSingh:
@jennydaman its happening on pf4.
@zrthxn , if the fetch on https://api.github.com/repos/${repo} this url returns a 404 then the page will crash as repoData.license will be undefined and accessing the name will throw an error

zrthxn:
I have tested that already by blocking the URL and seeing the behaviour, and there is error handling there to prevent a crash. Maybe it's not working exactly as I expected. I'll fix it asap. Thanks.

zrthxn:
The issue existed before I had added exception handling but that was fixed in [ebba6a8](https://github.com/FNNDSC/ChRIS_store_ui/pull/161/commits/ebba6a8f29fefe6bbe8df3a89c1170dbb34a2ec7) and [4499243](https://github.com/FNNDSC/ChRIS_store_ui/pull/161/commits/4499243ec53da375fbeaeafa9da6415fef23d6e7). 

But after testing by blocking the URL so that the fetch call doesn't work, the app does not crash and just gives a notification. 
However if the fetch call has a **404 status**, it doesn't throw an error and so the app crashes because the error handling doesn't take over.

PR #173, opened 2021-04-25
Fixed repository not found issue

zrthxn:
Fixes: #166 

The issue existed before I had added exception handling but that was fixed in [ebba6a8](https://github.com/FNNDSC/ChRIS_store_ui/pull/161/commits/ebba6a8f29fefe6bbe8df3a89c1170dbb34a2ec7) and [4499243](https://github.com/FNNDSC/ChRIS_store_ui/pull/161/commits/4499243ec53da375fbeaeafa9da6415fef23d6e7). 

But after testing by blocking the URL so that the fetch call doesn't work, the app does not crash and just gives a notification. 
However if the fetch call has a **404 status**, it doesn't throw an error and so the app crashes because the error handling doesn't take over.

### DIFF
--- a/src/errors/GithubError.js
+++ b/src/errors/GithubError.js
@@ -4,6 +4,7 @@ export class GithubAPIRepoError extends Error {
 
   /**
    * Create a new Github repo fetch error
+   * @param repo Name of repository
    * @param response Fetch API response
    */
   constructor(repo, response) {
@@ -20,6 +21,7 @@ export class GithubAPIRepoError extends Error {
 export class GithubAPIProfileError extends GithubAPIRepoError {
   /**
    * Create a new Github community profile fetch error
+   * @param repo Name of repository
    * @param response Fetch API response
    */
   constructor(repo, response) {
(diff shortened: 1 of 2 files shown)


jennydaman (commented):
Would you subclass the errors (say, `GithubReadmeFetchException`)?

zrthxn:
Will be done.

src/errors/GithubError.js
```diff
@@ -0,0 +1,42 @@
+export class GithubAPIRepoError extends Error {
+  statusCode;
+  response;
+
+  /**
+   * Create a new Github repo fetch error
+   * @param {*} response Fetch API response
```
jennydaman, line 7: Fix documentation

jennydaman (changes requested):
That's a detailed subclass! More than I would've done tbh.

src/errors/GithubError.js
```diff
@@ -0,0 +1,42 @@
+export class GithubAPIRepoError extends Error {
+  statusCode;
+  response;
+
+  /**
+   * Create a new Github repo fetch error
+   * @param response Fetch API response
```
jennydaman, line 7: Sorry for not being more specific earlier.

You should have a `@param` for every argument, in order. So please add one describing `repo`

jennydaman (changes requested):

src/errors/GithubError.js
zrthxn: Oh I added that param later so I forgot about that. Done now.

jennydaman (approved):

Issue #170, opened 2021-04-25
Search autocomplete selection with arrow keys


jennydaman:
Autocomplete was implemented in #154

Pressing either the <kbd>Up</kbd> or <kbd>Down</kbd> arrow keys should move the selection of which autocomplete suggestion to accept. The current behavior is that pressing <kbd>Down</kbd> just moves the cursor to the end.

@DivyanshiSingh do you want to take this issue?

DivyanshiSingh:
Yes, @jennydaman I would take this issue.

PR #174, opened 2021-04-25
Arrow key functionality in AutoComplete

DivyanshiSingh:
Fixes: #170 
1) Made a custom hook (useKeyPress hook) 
2) User can use the up and down key to select a particular item from the autocomplete list and can use enter key to search it.
@jennydaman please review and let me know the required changes.

### DIFF
--- a/src/components/Navbar/components/Search/Search.css
+++ b/src/components/Navbar/components/Search/Search.css
@@ -73,6 +73,11 @@
   padding: 5px 10px;
   font-weight: 500;
 }
+.active {
+  background: #e6e6e6;
+  cursor: pointer;
+  color: #3da1da;
+}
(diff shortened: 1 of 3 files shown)


jennydaman (approved):
It works so well!

EUGINELETHAL (commented):
@DivyanshiSingh @jennydaman some of the changes made here  make the search function not work as intended.
1)User is not able to search without authenticating
2.)I don't think the user should be restricted to inputting more than three characters so that the results  is returned , just prevent the default.I should be able to search just using two characters e.g 'pl'

jennydaman:
@EUGINELETHAL I am unable to reproduce (1)
I think (2) is a good feature to have, it prevents unnecessary requests. Two (or one) characters is just not enough information.

PR #176, opened 2021-04-26
PF4 plugins revamp

sonylomo:
Fixes issue #101 and #135
- Upgraded from class components to functional components.
- Upgraded from pf3 to pf4.
- Enable plugins sort.


jennydaman:
Hey @sonylomo there was a major overhaul to `Plugins.js` in #171 which unfortunately breaks most of this PR

sonylomo:
Yes, I've been working to resolve them.

src/components/Plugins/components/PluginItem/PluginItem.js
```diff
@@ -1,51 +1,48 @@
...
 
-
-class PluginItem extends Component {
-  renderStarButton() {
-    let name;
+const PluginItem = ({
+  isLoggedIn,
+  isFavorite,
+  onStarClicked,
+  id,
```
jennydaman, line 17: We should rename `id`i and `tittle` because they shadow HTML attribute names.

jennydaman (changes requested):

src/components/Plugins/components/PluginItem/PluginItem.js
sonylomo: I've changed them to `pluginId` and `pluginTitle`.

src/components/Plugins/Plugins.js
```diff
@@ -99,229 +117,164 @@ export class Plugins extends Component {
...
+                <LoadingContent
+                  width="135px"
+                  height="30px"
+                  left="1em"
+                  top="1.5em"
+                  bottom="1.5em"
+                />
+              </LoadingContainer>
+            )}
+            <DropdownButton
```
jennydaman, line 232: When `pluginList` is `null` (for instance, if the backend is unreachable) the page crashes.

jennydaman (changes requested):

jennydaman:
@sonylomo let me know on Slack if you need help rebasing.

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
```diff
@@ -1,249 +1,212 @@
...
-      password: '',
-      passwordConfirm: '',
-    };
-  }
-
-  handleChange(value, name) {
-    this.setState({ [name]: value });
-  }
+const DeveloperSignup = ({ store, ...props }) => {
+  let history = useHistory();
```
jennydaman, line 34: Why not `const history`?

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
```diff
@@ -1,249 +1,212 @@
...
+        fieldId="password-confirm"
+        validationState={
+          error.controls.includes("confirmation") ? "error" : "default"
+        }
+        helperText="Confirm your password"
+        inputType="password"
+        id="passwordConfirm"
+        fieldName="passwordConfirm"
+        value={passwordConfirm}
+        onChange={(val) => setPasswordConfirm(val)}
```
jennydaman, line 193: I don't think the arrow expression is necessary here.

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
```diff
@@ -1,249 +1,212 @@
...
 
-    this.setState({
-      loading: true,
-      error: {
-        message: '',
-        controls: '',
-      },
-    }, () => store.set('userName')(username));
+    setLoading(true);
+    setError({
```
jennydaman, line 87: I think we should flatten this object, i.e. replace with two state variables `errorMessage` and `errorControls`.
Moreover, the types are inconsistent -- `controls` is expected to be a list.

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
```diff
@@ -1,249 +1,212 @@
...
           });
         }
-        return resolve(e);
-      }
-      try {
-        authToken = await StoreClient.getAuthToken(authURL, username, password);
-      } catch (e) {
-        return reject(e);
+      } else {
+        setLoading(false);
```
jennydaman, line 120: Could we add an error message to the UI here too? Maybe that the network request failed?

src/components/Navbar/Navbar.module.css
```diff
@@ -0,0 +1,4 @@
+.login-button .other-button{
```
jennydaman, line 1: Is the filename `Navbar.module.css` correct?

jennydaman (changes requested):

src/components/Navbar/Navbar.module.css
sonylomo: It should be deleted.

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
sonylomo: Both error controls and messages have to be updated at the same time in the return statements like these:
`if (!password) {
      return setError({
        message: "Password is required",
        controls: ["password"],
      });
    }`

I don't think there's another way of doing this with the `error` object separated into 2 states. Unless you know another way 🤔.

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
sonylomo: Should I use notification or error handling on the input fields?

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
```diff
@@ -1,249 +1,222 @@
...
+        autofocus={!isTouchDevice}
+        onChange={(val) => setUsername(val)}
+        error={error}
+      />
+
+      <FormInput
+        formLabel="Email"
+        fieldId="email"
+        validationState={error.controls.includes("email") ? "error" : "default"}
+        helperText="Enter you email"
```
jennydaman, line 168: Would you fix this typo? "Enter your email"

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
```diff
@@ -1,249 +1,222 @@
...
-      }
-      try {
-        authToken = await StoreClient.getAuthToken(authURL, username, password);
-      } catch (e) {
-        return reject(e);
+      } else {
+        setLoading(false);
+        setError({
+          message: "Network request failed",
+          controls: ["username", "email", "password"],
```
jennydaman, line 130: In the event of a network request failure, we should probably have the `controls` array be empty, to convey that it is not an error of user input.

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
jennydaman: https://github.com/FNNDSC/ChRIS_store_ui/blob/ec5e30492c95f8c7933c8d6ef6b3563304988c61/src/components/Dashboard/Dashboard.js#L29

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
jennydaman: Do they have to be coupled or is it that it "just makes sense" to have them coupled?

Logically I think it'd be fine for us to allow an error message to be displayed without any fields turning red, or have fields turn red without an error message being displayed. While these situations wouldn't make sense (or at least are not friendly to the user) I think it makes development easier.

Though this is subjective, if you like how it is now or have your own idea, please do as you think best.

jennydaman (changes requested):

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
sonylomo: I think the arrow function is necessary for updating `PasswordConfirm` state with the input value.

src/components/FormInput/index.js
```diff
@@ -32,7 +33,7 @@ const FormInput = (props) => {
           {helperText}
         </FormHelperText> : null
       }
-      helperTextInvalid={error && error.controls.includes(fieldName) ? error.message: null}
+      helperTextInvalid={errorMessage && errorControls && errorControls.includes(fieldName) ? errorMessage: null}
```
jennydaman, line 36: What was the rationale for this change?
Could you describe to me what the purpose of `errorControls` is and why it's necessary?

Maybe we can simply have four state variables, `errorUsername`, `errorPassword`, `errorPasswordConfirm`, `errorEmail`, ... and no `errorControls`

sonylomo: Since the `errorControls` point to the field names i.e username, email, password, etc, I thought that it'd be reasonable to check if it's True. I also agree that it'd be better to separate the controls into 4 states.

jennydaman: Well `helperText` is one simple variable, could `helperTextInvalid` be coded up in the same way?

jennydaman: Simple as (theoretically) `helperTextInvalid=errorUsername`

> I can't reset the states of `errorUsername`, `errorPassword`, `errorPasswordConfirm`, `errorEmail` etc, after they've shown an error.

Why not? Could it be done in [`handleStoreLogin`](https://github.com/FNNDSC/ChRIS_store_ui/pull/176/files/5ea0f78b7498ed644c023af5398720330659295a#diff-bf517181be23941f250cd1ab597ce0e352b2d7aae0b14ace7b7317e51d92fd14R91)?

src/components/Plugins/Plugins.js
```diff
@@ -63,34 +112,18 @@ export class Plugins extends Component {
...
-   * @param error error message
-   */
-  showNotifications = (error) => {
-    console.error(error);
-    this.setState({
-      errorMsg: error.message,
-    })
-  }
+  const removePluginStar = (pluginId) => {
+    setPluginStar(pluginId, undefined);
```
jennydaman, line 125: a nitpick: typically `null` and `undefined` have the same effect, but in situations like this, it's better to set values as `null`. `null` denotes an empty value, and usually implies that the developer has set the variable's value to null intentionally. Sticking to such a practice makes it so that it's easier to identify logical errors when `undefined` pops up.

src/components/Plugins/Plugins.js
```diff
@@ -99,229 +132,204 @@ export class Plugins extends Component {
...
-  isFavorite(plugin) {
-    return this.state.starsByPlugin[plugin.id] !== undefined;
-  }
+  };
 
-  isLoggedIn() {
-    return this.props.store ? this.props.store.get("isLoggedIn") : false;
-  }
+  const isFavorite = (plugin) => {
+    return starsByPlugin[plugin.id] !== undefined;
```
jennydaman, line 188: If you do take the advice above and use `null` instead of `undefined`, be sure to also change the strict equality check here.

src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
```diff
@@ -1,249 +1,230 @@
...
+        />
+      )}
+      <Form onSubmit={handleSubmit} noValidate>
+        <p>{loading ? "Creating" : "Create"} a ChRIS Developer account:</p>
         <FormInput
           formLabel="Username"
           fieldId="username"
-          validationState={error.controls.includes('username') ? 'error' : 'default'}
+          validationState={
+            errorUsername.includes("username") ? "error" : "default"
```
jennydaman, line 155: I still do not understand why this approach is being used.

Here, `errorUsername`, `errorPassword`, etc..., are practically boolean variables. It do not think these are necessary.

My previous advice was to have four variables, `errorUsername`, `errorEmail`, `errorPassword`, `errorPasswordConfirm`. As we've discussed, having only one variable for `errorMessage` is poor program design because:

1. the view (in the context of the MVC pattern) should not care about what is being displayed. The view should allow for multiple error messages to be shown, e.g. both messages "Username is invalid" and "Password is empty" should be allowed to be shown at the same time. (If you think this is poor UX, then it is the model's responsibility to not set multiple error messages.)
2. in your implementation, `errorMessage` is serving one of four roles (showing error for username box, showing error for email box, showing error message for password box, showing error message for password confirm box). Which of these roles it serves is set by four other variables. So why not have one variable per role?

I think some of this confusion comes from how `src/components/FormInput/index.js` was originally implemented. Don't be scared to stray from original implementations (if you find them disagreeable).

Can you answer this question: "what is the purpose of the prop `errorControls` to `FormInput` (why do we need it)?"  In the 30 lines,  to me it seems like it serves no purpose.

This is also somewhere where documenting your code can be helpful. Every prop to a component should be documented. In well-documented code, I wouldn't have to ask such a question.

See if you can implement something like this. First, in pseudocode:

```
Does errorUsername have a value?
  If true, set validated="error" and helperText=errorUsername
  Else, set validated="default"
```

Suggested code:

```jsx
<FormInput
  formLabel="username"
  helperText="Enter your username"
  inputType="text"
  autoFocus={!isTouchDevice}
  onChange={(val) => setEmail(val)}
  errorMessage={errorUsername}
/>
```

Inside the code for `FormInput` it will decide: if `errorMessage` has a value, then pass the props `validated="error"` and `helperText=errorUsername` to `FormGroup`.

src/components/FormInput/index.js
```diff
@@ -19,23 +20,25 @@ const FormInput = (props) => {
     placeholder,
     className,
     defaultValue,
-  } = props;
+  } = props;
```
jennydaman, line 23: As aforementioned, I recommend adding documentation for all of these props. Any prop which seems to serve no purpose, or serve little purpose, should be removed. In particular, I find it strange that `fieldId`, `id`, and `fieldName` are usually set to the same value (like `"username"`)

Notice `id` is not being used.

I don't recommend `className`. Instead, spread the extra props on the parent.

```jsx
const FornInput = ({formLabel, fieldId, validationState, ...props}) => {
  return (<div ...props> [awesome code here] </div>)
}
```

src/components/Plugins/Plugins.js
```diff
@@ -99,229 +132,204 @@ export class Plugins extends Component {
...
         <div className="row plugins-row">
-          <PluginsCategories categories={categoryEntries}
-            onSelect={this.handleCategorySelect}
+          <PluginsCategories
+            categories={categoryEntries}
+            onSelect={handleCategorySelect}
           />
-          <div className="plugins-list">{pluginListBody}</div>
+          <div className="plugins-list">
+            {/* Plugin List Body*/}
```
jennydaman, line 302: Comments are good when informative, but I do not think this is informative.

jennydaman (changes requested):

src/components/Plugins/Plugins.js
jennydaman: @sonylomo you've marked this as resolved but did not make a change, does this imply that your software design choice is to use `undefined`?

sonylomo: @jennydaman, I hadn't commited the changes yet. I ran into another issue, for some reason the inputs are being registered as 2 values: a synthetic event and string (realized it after console logging). I've been trying to figure out why; maybe it's Patternfly. I've remade the Developer Signup component from scratch a couple of times but it still didn't work.

Issue #104, opened 2020-10-19
Generate project badge after upload


jennydaman:
e.g.

```
https://img.shields.io/badge/dynamic/json?color=blue&label=chrisstore.co&query=$.results%5B0%5D.version&url=https%3A%2F%2Fchrisstore.co%2Fapi%2Fv1%2F2%2Fplugins%2F
```

Zoey-dev:
Hi Jenny, I am an outreachy intern and i would like to work on this issue. Could you please assign to me?

Zoey-dev:
Hi @jennydaman I am an outreachy intern, please can you give more details to this issue.

jennydaman:
Many open-source repos on Github show badges which quickly summarize info about a project, and its current status.

e.g. https://github.com/FNNDSC/ChRIS_store#readme

Opportunities for providing a badge sometimes generate badges themselves. For example, the Github Actions tab has a "Create Status Badge" button in a drop-down menu, which opens a modal with copy-pastable Markdown code.

![image](https://user-images.githubusercontent.com/20404439/113522452-837ed280-956e-11eb-8cda-08f38e30ad6e.png)

We should automatically generate project badges for ChRIS plugins, which can be added to their READMEs. The purpose of the badge would be to

- signify that a repository is a ChRIS plugin
- maybe also show its version
- link to the info page and installation instructions at https://chrisstore.co

These badges can take one of two forms: a traditional plastic badge like

![plastic badge](https://img.shields.io/badge/dynamic/json?color=blue&label=chrisstore.co&query=$.results%5B0%5D.version&url=https%3A%2F%2Fchrisstore.co%2Fapi%2Fv1%2F2%2Fplugins%2F)

Or an App Store style badge like

![app store badge](https://raw.githubusercontent.com/FNNDSC/cookiecutter-chrisapp/master/doc/assets/badge/light.png)

This is inspired by other app stores, such as

<img src="https://f-droid.org/badge/get-it-on.png" alt="Get it on F-Droid" height="100"/>

![Chrome Web Store](https://user-images.githubusercontent.com/20404439/113522557-76aeae80-956f-11eb-843f-3c18788f06d8.png)
![Google Play](https://user-images.githubusercontent.com/20404439/113522578-aa89d400-956f-11eb-9377-5c5c1bb62c38.png)
![App Store](https://user-images.githubusercontent.com/20404439/113522581-bb3a4a00-956f-11eb-9737-902a43e2b581.png)

[...]

PR #177, opened 2021-04-28
Generate Readme badge

EUGINELETHAL:
What does this Pr do?
@jennydaman Fixes # https://github.com/FNNDSC/ChRIS_store_ui/issues/104#issue-724899335 Generates readmebadge for plugin
I opted for it to be on the dashboard so that it can only be viewable by the user who created it
Screenshot from 2021-04-29 01-17-37


src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.js
```diff
@@ -226,6 +228,10 @@ class DashPluginCardView extends Component {
                     {`${plugin.license} license`}
                   </div>
                 </div>
+                <div className="plugin-body-copy-url">
+                       
+                        <EmbedReadMe  text={`[![Chris Plugin](https://raw.githubusercontent.com/FNNDSC/cookiecutter-chrisapp/master/doc/assets/badge/dark-shiny.png)](https://chrisstore.co/plugin/${plugin.id})`} />
```
jennydaman, line 233: I would put the markdown code template inside the component instead of here as a prop. It makes more sense as component logic.

jennydaman: Also it would be cool if the `dark-shiny.png` and such can go in the ChRIS_store_ui's `src/assets` folder or `public` folder, then have the generated URL point to the ChRIS_store_ui's own host.

It's best to avoid using Github as a CDN. We can expect the ChRIS_store_ui to be served from a good static file server.

jennydaman (changes requested):

PR #182, opened 2021-05-04
Developer Components Upgrade to pf4

sonylomo:
Fixes issue #101 

- Changed class components to functional ones.
- Changed from pf3 to pf4
- Upgraded outdated react-bootstrap components.


src/components/Developers/components/DeveloperSignup/DeveloperSignup.js
```diff
@@ -1,249 +1,218 @@
...
-          </Button>
-        )}
-        {loading && <span className="developer-signup-creating">Creating Account</span>}
-      </Form>);
-  }
-}
+
+  };
+
+  const disableControls = loading;
```
jennydaman, line 135: We should take this out

jennydaman, line 135: We should take this out

src/components/Navbar/Navbar.js
```diff
@@ -34,6 +34,7 @@ const navLinks = [
  * Conditionally renders a list of links into a <Nav>.
  */
 const Navigation = ({ store }) => {
+  const location = useLocation()
```
jennydaman, line 37: This is a good change. It might make prerendering or SSR easier too.

src/components/Plugins/Plugins.js
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Client from '@fnndsc/chrisstoreapi';
-import { DropdownButton, MenuItem } from 'react-bootstrap';
+import { DropdownButton, Dropdown } from 'react-bootstrap';
```
jennydaman, line 4: Do you think we can use Patternfly components here, replacing `react-boostrap`?

sonylomo: Yeah I can.

src/components/Navbar/Navbar.js
sonylomo: Thanks.

sonylomo:
I've used id to style the `sort by` button to remain on the right side of the page. I tried using classname but I had to add `!important` on every style for it to work. I hope it's ok.

src/components/Plugins/Plugins.css
```diff
@@ -29,6 +29,16 @@
...
 }
 
+#sort-by-dropdown{
+  margin: 1.35em 1em ;
+  position: absolute ;
+  right: 0 ;
+}
+
+#sort-by-dropdown-item{
+  list-style: none !important;
```
jennydaman, line 39: This shouldn't be necessary, it should be handled by Patternfly. We should investigate. Maybe transpile config is wrong, maybe an update would help?

jennydaman (changes requested):

src/components/Plugins/Plugins.css
sonylomo: I've tried to update but it hasn't changed.

jennydaman: Found it! `patternfly.css` is being left out. Maybe we should add the line below to `index.js`?

```javascript
import '@patternfly/patternfly/patternfly.css';
```

sonylomo: cool

jennydaman (approved):

Issue #151, opened 2021-04-15
Use CSS modules


jennydaman:
https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/

somya51p:
@jennydaman  Can you give a brief on this issue. Can I work on it?

jennydaman:
The [pf4](https://github.com/FNNDSC/ChRIS_store_ui/tree/pf4) branch is a piece-by-piece rewrite of the *ChRIS_store_ui*. We want to make sure the code quality and maintainability is best.

In this issue, we want to evaluate the feasibility of using CSS modules. CSS modules are explained in the link above. Then, we want to refactor the project to use CSS modules.

jennydaman:
There are currently 278 lines of code containing the word `className` in the codebase, so this is not a small issue.

somya51p:
@jennydaman I wish to work on this issue. Can you please assign me the same.

jennydaman:
Pointed out by @DivyanshiSingh there are new bugs introduced by #162, @somya51p could you take a look?

jennydaman:
![screenshot](https://user-images.githubusercontent.com/20404439/117501063-901a9000-af4b-11eb-9076-bc4d2c0cd02b.png)

jennydaman:
@sonylomo fixed some of it in #182 which I just merged.

![screenshot](https://user-images.githubusercontent.com/20404439/117503068-72025f00-af4e-11eb-9a4c-917df4e3201c.png)

somya51p:
@jennydaman Sure, I will look into it.

DivyanshiSingh:
@jennydaman I think we should have taken the changes piece by piece for every component and not at once. As now I can see there are some unwanted changes done to the previous merged PRs and also css bugs.
![Screenshot 2021-05-08 at 12 38 04 PM](https://user-images.githubusercontent.com/53593872/117530461-f14d6c80-affa-11eb-8906-a1ada35509a9.png)


![2021-05-08-2](https://user-images.githubusercontent.com/53593872/117530590-9a946280-affb-11eb-9a7d-fea547f57584.png)


![2021-05-08-1](https://user-images.githubusercontent.com/53593872/117530593-9c5e2600-affb-11eb-9b32-d23e829d04fc.png)


![2021-05-08](https://user-images.githubusercontent.com/53593872/117530594-9d8f5300-affb-11eb-9172-6b24f52a9b9f.png)

[...]

PR #190, opened 2021-05-08
Resolved bugs caused by CSS modules

somya51p:
@jennydaman This PR resolved the bugs caused while shifting to `CSS modules` under issue #151. That was because `CSS modules` are _CSS files in which all class names and animation names are scoped locally by default_
 **If I am missing out on something, please let me know.**

Screenshot from 2021-05-08 21-13-48


jennydaman:
I haven't dug into how `react-scripts` wraps `webpack` to bundle regular CSS vs CSS modules.

If you are interested in this issue, I can do that and work with you to figure out how we can mix regular CSS and CSS modules to do an incremental upgrade.

Another idea is, we could use regular expressions and scripting. An automatic approach would be generalized, so we can be more sure that it's correct. It would also facilitate converting kebab-case to camelCase which will be easier for developers to work with.

jennydaman:
As a proof of concept, a basic regex would look like `(?<=className=)'.+'`. Here, we are finding the string between quotation marks which follows `className=`. For simplicity, the regex does not account for whitespace nor single-vs-double quote (these would be easy features to add).

https://regexr.com/5sq5b

We take the result and split the result on whitespace to obtain a list of CSS classes. We convert the name to camelCase and rewrite the string to use CSS modules. We can also do an automatic assertion for whether or not the class is found in any imported CSS file.

somya51p:
@jennydaman Yeah sure, I would like to work on that.

somya51p:
@jennydaman Should I start working on converting the name to camelCase and rewrite the string to use CSS modules. Also, would it be better to convert classes piece-by-piece? Please suggest.

jennydaman:
@zrthxn just brought up the suggestion we work with SCSS instead (which has the benefits of CSS modules).
Let's think about it.

jennydaman:
Either way, before we do this we should first:

1. merge pf4 into master
2. create a new branch for unstable refactors

jennydaman:
We will continue this discussion in the respective issue threads.

Issue #203, opened 2021-05-25
Change all React .js files to .jsx


zrthxn:
2.0 milestone
[Refactoring]

PR #200, opened 2021-05-20
Addressing issue no #198

praguru14:
Have separated isTouchDevice form DeveloperSignup.js

Issue #198, opened 2021-05-20
isTouchDevice() should be moved to a separate file


jennydaman:
https://github.com/FNNDSC/ChRIS_store_ui/blob/0d150714d6dada40831251796d49e735f6fa756c/src/components/Developers/components/DeveloperSignup/DeveloperSignup.js#L17

PR #197, opened 2021-05-14
CopyToClipboard: tooltip upgraded to pf4

DivyanshiSingh:
Fixes: #101 
1- Used CopyToClipboard feature from react-copy-to-clipboard.
2- Upgraded tooltip from pf3 to pf4.

PR #207, opened 2021-05-26
Major Refactoring

zrthxn:
Closes: #203 
Closes: #101 
Closes: #197
Fixes: #198  
Closes: #200 

Summary
- Removed almost all PF3 components.
- Change all React `.js` files to `.jsx`
- Updated Jest snapshots
- CopyToClipboard: tooltip upgraded to pf4
- Move isTouchDevice to separate file


zrthxn:
@jennydaman Completed PF3 removal.

src/components/Developers/components/CopyToClipboard/CopyToClipboard.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Clipboard from 'react-clipboard.js';
+import { Tooltip } from '@patternfly/react-core';
+import { CopyIcon } from '@patternfly/react-icons';
+import './CopyToClipboard.css';
+
+const CopyToClipboard = ({ clipboardText }) => (
+  <d>
```
jennydaman, line 9: Did you mean `<a>`?

jennydaman (changes requested):

jennydaman:
Crashes after login.

> Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
>
> Check the render method of `DashPluginCardView`.

src/components/Developers/components/CopyToClipboard/CopyToClipboard.jsx
zrthxn: Oops, that was supposed to be a div!

zrthxn:
> Crashes after login.
> 
> > Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
> > Check the render method of `DashPluginCardView`.

Thats because of the `MessageDialog` component. It'll have to be commented out for the time being until an equivalent works. I've told @sonylomo about it to get something working asap.

jennydaman:
Oh yeah, I was just digging at that. So what was the way forward with fixing the `MessageDialog` component, what's @sonylomo's involvement?

zrthxn:
Right now I dont think she's gotten to it. I'll use the Modal in PF4 to build a replacement.

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.jsx
```diff
@@ -3,13 +3,15 @@ import PropTypes from "prop-types";
...
 import {
-  Col,
   EmptyStateAction,
   EmptyStateInfo,
   FieldLevelHelp,
   MessageDialog
-} from "patternfly-react";
-import { CardTitle, CardBody, Card, DropdownItem, Dropdown, KebabToggle, GridItem, Grid, Form } from "@patternfly/react-core";
+} from "@patternfly/react-core";
+import { CardTitle, CardBody, Card } from "@patternfly/react-core";
```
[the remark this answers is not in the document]
zrthxn: Done

jennydaman:
The project has a different error now. When clicking on a plugin card in `/dashboard` I get

> Unhandled Rejection (TypeError): pluginData is undefined
> componentDidMount
> src/components/Plugin/Plugin.jsx:60

I merged master to resolve merge conflicts. This did not change any of the files.
To be certain, this bug exists in e350365afe48643829f329ee4429cf30e36c41bf

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.jsx
```diff
@@ -0,0 +1,301 @@
...
+      
+    const pluginCardBody = plugins.map((plugin, index) => {
+      const creationDate = new RelativeDate(plugin.creation_date);
+      const applicationType = DashApplicationType(plugin.type);
+      return (
+        <GridItem sm={12} md={6} key={plugin.name}>
+          <Card style={{ height: '100%' }}>
+            <CardHeader>
+              <CardTitle>
+                <Link to={`/plugin/${plugin.name}`}>{plugin.name}</Link>
```
jennydaman, line 161: I found the error.

Here, the URL is being set to something like `chrisstore.co/plugin/pl-example`.

In [`src/components/Plugin/Plugin.jsx:fetchPluginData()`](https://github.com/zrthxn/ChRIS_store_ui/blob/109f28357adee3e58b80f9c8ef546ef33b0a5c13/src/components/Plugin/Plugin.jsx#L121) the function is expecting an integer (plugin ID), not plugin name.

Three comments: first, the quickest fix here would be to set the URL to something like `chrisstore.co/plugins/7` instead.

Second, I think having plugin name instead of plugin ID in the URL is a better UX. This would require some thinking and an overhaul. Currently, the page `chrisstore.co/plugins/7` shows details about one version of a plugin. We want to create a page `chrisstore.co/plugins/pl-example` which shows details about all plugins with the same name `pl-example`, and perhaps has a "Versions" tab which links to specific plugins by ID.

Third, regardless, `Plugin.jsx` needs refactoring. In `fetchPluginData()` the function catches network exceptions but still goes on to return `undefined` and pollute the state. On network error, the exception should be caught then rethrown.
In `componentDidMount()` state management is done messily. We should change `fetchPluginData()` into a stateless, deterministic function.
In situations where the state does not have the plugin data, we should not fetch the plugin data. Instead we should wait for the state to change.

jennydaman (changes requested):

src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.jsx
zrthxn: Sorry I'm late on this. 

For the first two comments, i also think having the name in the url is better. And it'd be good to have it show all versions with the name and a specific version with the ID. The fetching function can check for an integer (or hex in future) ID and if that fails then treat it as a name. 

For the third comment, yes I agree with that. I had done that same thing before, with re-throwing the exceptions but then we had a discussion about how we should only catch if we can handle it then and there.

jennydaman: Would you implement the quick fix, that is, have the plugin ID in the URL?
I would like to merge this PR just to unblock other tasks, such as #209 
Also, please create detailed issues to start the discussion on how we can organize routing and information to support URLs with plugin names.

jennydaman: > The fetching function can check for an integer (or hex in future) ID and if that fails then treat it as a name.
I disagree with "guessing" union types. It is safer to be explicit.

> or hex in the future
What do you mean?

zrthxn: Okay I've done the quick fix for now, the rest of the changes will be in the other PR (like having the name in the URL).

> I disagree with "guessing" union types. It is safer to be explicit.

Then the URL can have something like `/plugin/plugin-name` for the named metas that gets all the versions and `/plugin/p/:id` for fetching one specific one by ID.

> What do you mean?

Maybe with growing plugins integer IDs could be handled as hex numbers, but thats way ahead in the future.

jennydaman: > Then the URL can have something like `/plugin/plugin-name` for the named metas that gets all the versions and `/plugin/p/:id` for fetching one specific one by ID.

Sounds better.

> Maybe with growing plugins integer IDs could be handled as hex numbers, but thats way ahead in the future.

Yeah, way ahead into the future.

jennydaman:
Regressions

#### "Sort By" menu options leak out

image

#### Plugin details sidebar overlaps

image

#### Dashboard (after sign-in) CSS flaws

Fix spacing next to revisions pane

image

#### Teammates list are plugins again

This seems new, could you explain what is the intention here?

image


#### Get a UI notification error "Failed to get plugin stars" when looking at plugin details while logged in

jennydaman:
#209 inherits from this branch

Issue #199, opened 2021-05-20
Replace shortened links with real links


jennydaman:
https://github.com/FNNDSC/ChRIS_store_ui/blob/0d150714d6dada40831251796d49e735f6fa756c/src/components/Developers/components/Instructions/Instructions.js#L10

- the link here is outdated
- URL shorteners do not respect privacy

This needs to be changed to https://github.com/FNNDSC/cookiecutter-chrisapp/wiki/Quickstart

jennydaman:
We should replace all of them.

```bash
$ rg bit.ly
src/components/Developers/components/Instructions/Instructions.js
10:        <a className="instructions-source" href="http://bit.ly/2KghHdY">[source]</a>
36:                    href="http://bit.ly/2Iih52m"
45:                    href="http://bit.ly/2KghHdY"
141:              <a href="http://bit.ly/2yzbRzF">simple <strong>fs</strong> plugin</a>
143:              <a href="http://bit.ly/2KbHosS">simple <strong>ds</strong> plugin</a>
156:              consult the <a href="http://bit.ly/2ltTJ0w">wiki</a> to learn how to register it to
```

SugatoBagchi:
Hi,

I would like to work on this issue. Can you please assign it to me?

Thank you.

PR #231, opened 2021-07-22
Bug fixes and Plugin Versions content

zrthxn:
Fixes: #199 
Fixes: #230 
Screenshot from 2021-07-23 01-59-55

## Versions Content 
Add list of versions to content section and add choice of which version to install in the install button.

Screenshot from 2021-07-23 01-57-55
Screenshot from 2021-07-23 01-57-46


src/components/Plugin/Plugin.jsx
```diff
@@ -67,7 +68,7 @@ export class Plugin extends Component {
     if (this.isLoggedIn()) {
       return this.isFavorite() ? this.unfavPlugin() : this.favPlugin();
     }
-    return Promise.resolve();
+    return this.showNotifications(new Error('You need to be logged in!'))
```
jennydaman, line 71: I wonder what the old `Promise.resolve()` was doing.

`this.showNotifications` doesn't return anything, I don't think you should use `return` in `onStarClicked`.

As a future idea, a better design here might be to split this class into `AbstractPlugin`, `LoggedInPlugin`, and `AnonymousPlugin`. Dynamic dispatch would help the code stay clean and have fewer `if isLoggedIn`)

src/components/Plugin/components/PluginBody/PluginBody.jsx
```diff
@@ -81,7 +80,38 @@ const PluginBody = ({ pluginData }) => {
       fetchRepo();
   }, [fetchReadme, fetchRepoData, pluginData.public_repo, showNotifications]);
 
-  const versionString = (v) => `Version ${v}${v === pluginData.version ? ' (This)' : ''}`;
+  const InstallButton = () => {
+    if (pluginData.version) 
+      return <>
+        <p><b>Version { pluginData.version }</b></p>
+        <ClipboardCopy isReadOnly>
+          { `${process.env.REACT_APP_STORE_URL}plugins/${pluginData.id}/` }
```
jennydaman, line 88: The plugin URL should be retrieved from the API response, not constructed via string template/concatenation.

If possible, avoid reading `proces.env` for any reason other than to create the client object.

jennydaman (changes requested):

src/components/Plugin/Plugin.jsx
zrthxn: The old `Promise.resolve()` was only because we have to return something at the end of an arrow function, according to the Airbnb style guide we've implemented. For the same reason I'm returning `this.showNotifications` even though it returns `void`.

jennydaman: Ah, yeah I see that the style enforces that a function must either always return something or never return something. It's better for this void function to never return anything. Your latest changes are good.

zrthxn:
@jennydaman After this is merged, the master branch can be deployed.

jennydaman (changes requested):
I noticed that the screens for "plugin meta" (`/plugin/pl-name`) v.s. "plugin" (`/p/14`) are the same. This is okay for now but we will need to make them distinct in the future.

I think I said this before, the "Install to ChRIS" url should be something from the backend API's data. It should not be built from string concatenation/template. Aside from that, the template you are using is not correct for `/p/<N>` routes.

image

jennydaman:
Also note that all backend API handles must end in trailing slashes. However, you shouldn't have to worry about this if you are using the client correctly, that is, avoid concatenating URLs and instead use the URLs as they are returned by the backend.

zrthxn:
The URL was only being built for the `/plugin/pl-name` route. Otherwise I am using the URL from the response to show in the box. I'll fix this right now.

src/components/Plugin/components/PluginBody/PluginBody.jsx
```diff
@@ -81,7 +81,37 @@ const PluginBody = ({ pluginData }) => {
...
+          {pluginData.url}
+        </ClipboardCopy>
+      </>
+
+
+    if (pluginData.versions)
+      return <>
+        <p><b>Version { pluginData.versions[0].version }</b></p>
+        <ClipboardCopy isReadOnly>
+          {pluginData.url + pluginData.versions[0].id}
```
jennydaman, line 97: String concatenation is still being used here to "guess" what the correct URL is here.
It's not correct because there is no trailing slash.

But also, shouldn't the `pluginData.versions[0]` object have a `url` field?

jennydaman (changes requested):

src/components/Plugin/components/PluginBody/PluginBody.jsx
```diff
@@ -81,7 +81,37 @@ const PluginBody = ({ pluginData }) => {
...
+        </ClipboardCopy>
+        <br />
+        <ExpandableSection toggleText="More Versions">
```
jennydaman, line 100: Please add a conditional so that this `ExpandableSection` is not shown when there is only one version.

zrthxn: Done

src/components/Plugin/components/PluginBody/PluginBody.jsx
zrthxn: No the URL field exists on the PluginList object, not on the Plugin, and so not on the versions array. Adding that is not difficult but it adds yet another loop to the rendering time.

zrthxn: But the trailing `/` is added by the backend, we aren't guessing it. The format of all URLs returned by the API has that trailing `/`

jennydaman: Sorry, I was guessing based on what I know the JSON responses are like. Maybe the `@fnndsc/chrisstoreapi` should include the url, or there's a different endpoint to hit?
For instance, `https://chrisstore.co/api/v1/14/plugins/` supplies the `href` field which has the value `https://chrisstore.co/api/v1/plugins/14/`
https://hoppscotch.chrisproject.org/?method=GET&url=https://chrisstore.co&path=/api/v1/14/plugins/&headers=%5B%7B%22key%22:%22accept%22,%22value%22:%22application/vnd.collection+json%22,%22active%22:true%7D%5D

zrthxn: I'm using the `getPlugins` method on the Store API client, with 
```
getPlugins({
  // searchParams
  name
})
```

This should be the right thing to do, [according to the docs](https://fnndsc.github.io/fnndsc/chrisstoredoc/class/src/client.js~Client.html#:~:text=plugin%20id%20exactly%20with%20this%20number-,searchParams.name,match%20plugin%20name%20containing%20this%20string,-searchParams.name_latest)

jennydaman: What I am saying is, I think the `PluginMeta.getPlugins` method should be used not `Client.getPlugins`.

The former might have `url` data but not sure, I'll try in a bit.

src/components/Plugins/Plugins.jsx
```diff
@@ -363,21 +363,29 @@ export class Plugins extends Component {
...
           window.scrollTo(0,0);
-          const { name: query } = routeProps.match.params
-          if (pluginList.has(query)) {
-            const plugin = pluginList.get(query);
+          const { name } = routeProps.match.params
+
+          if (loading) return <LoadingPluginItem />
+
+          /**
+           * When the user opens this route from `/plugins`, the pluginList Map
```
jennydaman, line 373: This block of comments do not seem to be describing to following block of code. The `name_exact` fallback is happening on [line 150](https://github.com/FNNDSC/ChRIS_store_ui/pull/231/files#diff-21ceed2791318acc4d4ba806f53e82b4d3c6013e3a5de8373c219aa9c4979e00R150).

jennydaman (changes requested):
I finally got around to understanding how this all works, and a few notes.

The `<Switch>` in `Plugins.jsx` is nested under the component exported by `Router.jsx` right? Having a nested router which manipulates routes at the same level (here, both are manipulating top-level routes) is highly prone to errors. I suggest refactoring such that the application only has one top-level `<Switch>`.

Regarding `<Plugin>`: optional props are okay, and here `pluginData` is optional. But self-fulfilling optional parameters is poor design. When arguments are not given to an object (component), the object should either show a default or do nothing (component is blank), it should not try and fulfill its own arguments (props). This is especially true in the world of React because you're blurring the line between props and state, where props are not mutable by the component itself.

On top of the "self-fulfilling parameter" the [if-else of componentDidMount](https://github.com/FNNDSC/ChRIS_store_ui/pull/231/files#diff-59266294bc4b1481331191fa3b81a81ff6b60f0f3f71995098fef3e6eb2a21ffR49-R57) causes further branching of logic leading to high code complexity.

In OOP there is a concept "inversion of dependency," where it is the client's responsibility to coerce data to the object's interface. In React speak, the parent component in `Plugins.jsx` (more specifically, whichever `Switch > Route` is the parent to `<Plugin>`) should be responsible for wrangling its data to match the expected props for `<Plugin>`. In effect, you should move the complicated network requests and data (`pluginData`) coercion code within the execution tree of `Plugin.componentDidMount`  (so also consider `fetchPluginVersions` and `fetchPluginData`) to its parent (which in your implementation, the parent is defined in `Plugins.jsx`). The goal would be for `Plugin.jsx` to define a stateless component which does nothing more than display the data its given, without trying to modify nor create data in any way. It would be the sole responsibility of `Plugins.jsx` to provide the correct data. This responsibility of "providing the correct data" should happen in one place and one place only. If you want to support the `/plugins` to `/plugin/pl-name` data saving feature, and you want `<Plugin>` to be nested under `<Plugins>`, then `<Plugins>` should be responsible for providing the correct data to `<Plugin>` _in all cases_, and `<Plugin>` should be a "dumb" stateless component which operates with what it is given and does nothing else.

src/components/Plugins/Plugins.jsx
```diff
@@ -79,8 +71,10 @@ export class Plugins extends Component {
   }
 
   /**
-   * Re-fetch the list of plugins if the input was changed
+   * Re-fetch the list of plugins if the input was changed 
    * in the NarBar's search bar.
+   * @param {*} pluginId
```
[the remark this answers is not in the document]
zrthxn: Those were already there, I didn't add that, even though since I changed it, git shows that I added it.

zrthxn:
@jennydaman Okay so to understand better 

1. Why are nested routes like this a bad idea? I'd have thought that it's necessary to do this to have different components rendered by route. I've used this in the Chris ui app too. So I don't know how it'd cause problems. I'd need to know if I am to fix it.

2. We aren't self-fulfilling props exactly. Yes we're fetching data which wasn't provided in props but we're using the state to do this. In any case, if we were to use plugin.jsx to just render then its parent, Plugins.jsx would need to also handle fetching by ID which would mean that it'd have to either guess what type of param is provided (either string name or number ID). 

3. The if-else can be made to just be if. The else part can be common to all renders, meaning the plugin page would show the versions tab even if I opened a plugin by ID (populated by other versions). 

I think if one component was to do all the data coersion it would first mean that that component would be too big and certain methods in that component would only be used when children are rendered, not when it's own content is rendered. So that would probably go against separation of concerns. 
Moreover, if the same reasoning was to be applied to Plugin.jsx and PluginBody.jsx then the body should only render and the Plugin.jsx component should fetch the readme and all of those things which are currently done in PluginBody.jsx. Would that be right?

src/components/Plugins/Plugins.jsx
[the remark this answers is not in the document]
jennydaman: Oh okay. It's outdated, would you fix?

zrthxn:
Will be done

jennydaman:
1. In terms of tradeoffs to make, having nested routers isn't that bad. We should keep your solution if you think it works best.

Here is the reason I would argue it is convoluted (not ideal):

The `<Switch>` is a handler of sorts, which responds to when the location is updated (`/plugins` to `/plugin/pl-z2labelmap`). Let's abstract the parts here:

- URL's location as `SourceA`, 
- `<Switch>` in `Switch.jsx` as `HandleA`
- `<Switch>` in `Plugins.jsx` as `HandleB`
- The HTML as `SinkA`

When there is one `<Switch>` which routes top-level locations, we have the following

```
(SourceA) --[listened to by]--> <HandleA> --[updates]--> (SinkA)
```

Let's start on common ground and agree that having only one `<Switch>` is a paradigm with zero complexity.

What about when we have two Switches, which listen on the same resource, and affect the same resource?

```
(SourceA) --[listened to by]--> <HandleA> --[updates]--> (SinkA)
   \                                                        /
    --[listened to by]--> <HandleB> --[updates]-------------
```

In this situation, complexity is increased because `HandleB` and `HandleA` share the same source and sink. If `HandleA` updates `HandleB`, it is now possible for `HandleB` to enter an undefined state.

jennydaman:
> we're fetching data which wasn't provided in props but we're using the state 

2. Exactly what I am criticizing, you are blurring the line between props and state.

If you weren't given a certain piece of data as a prop, you shouldn't be trying to find the data yourself and then storing it as a state.

If a category of data is being represented by a prop, then the component should consider that piece of data immutable.

This is a design choice that you must go all-or-nothing: do you want

A) `Plugin` to be responsible for fetching its own `pluginData`, or
B) `Plugin` to be "dumb," have `pluginData` be a prop, where `Plugin` is incapable of fetching itself, and it's the sole responsibility of `Plugin`'s parent to supply it with `pluginData`?

Hint: if you are interested in the feature where data from `/plugins` can be reused when navigating to `/plugin/pl-z2imageslices`, go with the latter design option.

jennydaman:
> So that would probably go against separation of concerns.

You are trying to use the code for showing details about a plugin for _both_ showing plugin details or plugin meta details, where plugin != plugin meta. So we've already bit the bullet (bitten? bullet has been bit?)

jennydaman:
The most ideal but time-consuming way forward would be to develop two separate screens for plugin vs plugin meta details. If you don't want to/don't have time for that. then we have to do some sort of data cohesion somewhere, and we are debating where is best for this code to be. My argument is that the code for data cohesion should not exist in the `Plugin.jsx` file because of dependency inversion. It should be the role of its parent, which is coded up in `Plugins.jsx`. If you think the amount of code in `Plugins.jsx` is too much, then create a `PluginMetaAdapter.jsx` which is imported by `Plugins.jex`

jennydaman:
3. The if-else can be made to just be if.

It's not a big deal, I am just pointing out this `if` statement because it indicates that the interface to `Plugin.jsx` is a union data type.

I would prefer it if the if/else can be moved outside of `Plugin.jsx` so that its interface can be static, or something that's at is simpler.

zrthxn:
The nested `Switch` was removed and each component only renders itself now.
The following are word descriptions of what they do
- Plugins.jsx  => _Shows a list of all plugin metas_
  **params** { None }
  **returns** { List of Plugin Metas }

- PluginMeta.jsx  => _Given name, shows a plugin meta and its other versions_
  **params** { Plugin name from URL }
  **returns** { Plugin Meta }

- Plugin.jsx  => _Given ID, shows a plugin and its other versions_
  **params** { Plugin ID from URL }
  **returns** { Plugin }

Their behavior is now predictable.

src/components/Plugin/Plugin.jsx
```diff
@@ -65,9 +71,13 @@ export class Plugin extends Component {
...
     if (this.isLoggedIn()) {
-      return this.isFavorite() ? this.unfavPlugin() : this.favPlugin();
+      if (this.isFavorite()) 
+        this.unfavPlugin();
+      else 
+        this.favPlugin();
     }
-    return Promise.resolve();
+    else
+      this.showNotifications(new Error('You need to be logged in!'))
```
jennydaman, line 80: [Nitpick] I suggest rephrasing this message, something like

> Login to vote for this plugin as a favorite.

https://www.patternfly.org/v4/ux-writing/error-messages

jennydaman (approved):

src/components/Plugin/Plugin.jsx
zrthxn: Done

PR #241, opened 2021-10-08
Bump dependencies so we can build w/o resolutions

### LINKED ISSUE #238: Bump dependencies so we can build w/o resolutions
Upgrade dependency versions so that we can build without the hotfix https://github.com/FNNDSC/ChRIS_store_ui/pull/237

OladapoAjala:
@jennydaman this PR fixes #238

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,9 +53,6 @@
     "table-resolver": "^4.1.1",
     "undux": "^5.0.0-beta.18"
   },
-  "resolutions": {
-    "react-scripts/postcss-preset-env/postcss-custom-properties": "^10.0.0"
-  },
   "browserslist": {
     "production": [
       ">0.2%",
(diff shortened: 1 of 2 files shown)


zrthxn:
You're not supposed to edit the `yarn.lock` file to do this. This is because your changes will get overwritten whenever a new package is installed. The lock file is meant to be just for the package manager to handle module resolution. It even says this on the top of the file.

You should instead change the versions of the `react-scripts` package to test and see what works.

OladapoAjala:
That was my first approach, I upgraded react-scripts to v4. However, there are other libraries that also depend on postcss-values-parser v2.0.1 hence it didn't get updated. 

Except I upgrade all the other libraries which have a common sub-dependency with react-scripts then I wouldn't have to edit the yarn.lock file.

Issue #246, opened 2021-10-20
Search results bug


jennydaman:
Searching for a plugin which does not exist shows "-1 plugins found"

https://next.chrisstore.co/plugins?q=doesnotexist

![screenshot](https://user-images.githubusercontent.com/20404439/138033015-3ff33b3e-1428-4d6c-9fe3-6a3b0e5a3bc9.png)

mairin:
Is this one fixed? Can I close it?

somya51p:
@jennydaman Is this issue still open to work on?? if so please assign it to me.

somya51p:
@mairin @jennydaman just found a minor bug in the pagination part while searching plugins
![Screenshot from 2022-03-30 19-37-29](https://user-images.githubusercontent.com/64978837/160854525-a070cd12-33d1-4a3e-a761-d0d4cc09b6fe.png)

Kindly assign me this issue so that I can fix it and create a PR

PR #268, opened 2022-04-05
fixed plugin search issue under pagination section

somya51p:
@jennydaman Please review the PR and suggest changes if any. It resolved issue #246 from my end. 

Screenshot from 2022-04-05 13-10-47

### DIFF
--- a/src/components/Plugins/Plugins.jsx
+++ b/src/components/Plugins/Plugins.jsx
@@ -369,7 +369,7 @@ export class Plugins extends Component {
     )
 
     const pluginsCount=plugins.totalCount > 0 ? plugins.totalCount : 0;
-    
+    const pluginsLast = Math.min(paginationOffset + paginationLimit, pluginsCount);
 
     return (
       <article>
@@ -411,22 +411,18 @@ export class Plugins extends Component {
                               bottom="1.5em"
                             />
                           </LoadingContainer>
-                        ) : (
+                        ) : ( pluginsCount===0 ?
+                            <span style={{ color: 'gray', fontSize: '1.5em', margin: '1em 0' }}>
+                            <p style={{ fontSize: '1.25em', margin: '0', color: 'black', fontWeight: '600' }}>
+                                No plugins found
+                            </p>
+                          </span>
+                            :
                           <span style={{ color: 'gray', fontSize: '1.5em', margin: '1em 0' }}>
                             <p style={{ fontSize: '1.25em', margin: '0', color: 'black', fontWeight: '600' }}>
                                 {pluginsCount} plugins found
                             </p>
-                            Showing {paginationOffset + 1} to {' '}
-                            {
-                              // eslint-disable-next-line no-nested-ternary
-                              (paginationOffset + paginationLimit > plugins.totalCount) ?
-                                plugins.totalCount
-                                :
-                                (paginationOffset > 0) ?
-                                  paginationOffset
-                                  :
-                                  paginationLimit
-                            }
+                            Showing {paginationOffset + 1} to {pluginsLast}
                           </span>
                         )
                       }


jennydaman:
Hey @somya51p sorry for not noticing your comment on issue #246.

Could you explain to me why this bug is happening? Your fix seems like a band-aid solution.

somya51p:
> Hey @somya51p sorry for not noticing your comment on issue #246.
> 
> Could you explain to me why this bug is happening? Your fix seems like a band-aid solution.

@jennydaman For the pagination offset it was **showing 1 to -1**, which is irrelevant, so I have just tried to replace it according to the `plugins.totalCount` as totalCount = -1 is coming from the backend side as discussed in issue #246 earlier.

jennydaman:
Ah, thanks @somya51p, Please forgive my bad memory.

I do remember this discussion, I'm not sure but I think I might've said somewhere that it'd be better to fix this at the level of the library `@fnndsc/chrisstoreapi`? Or perhaps at a higher level, with an explanation? Ideally we would like to separate application logic from display logic.

I also see this line of code which should probably be used:

https://github.com/FNNDSC/ChRIS_store_ui/pull/268/files#diff-21ceed2791318acc4d4ba806f53e82b4d3c6013e3a5de8373c219aa9c4979e00R371

somya51p:
@jennydaman I get your point, in case of no result it should obviously show something like - 
Screenshot from 2022-04-07 17-48-50

and would this code be satisfactory for the above case? -

Screenshot from 2022-04-07 17-52-47

jennydaman:
I don't understand the context of your screenshot, would you generate a patch instead? To do so, you can run

```
git diff > changes.patch
```

Then paste the contents of changes.patch here.

somya51p:
> I don't understand the context of your screenshot, would you generate a patch instead? To do so, you can run
> 
> ```
> git diff > changes.patch
> ```
> 
> Then paste the contents of changes.patch here.

@jennydaman Please have a look!
```
diff --git a/src/components/Plugins/Plugins.jsx b/src/components/Plugins/Plugins.jsx
index 1748db8..51047d2 100644
--- a/src/components/Plugins/Plugins.jsx
+++ b/src/components/Plugins/Plugins.jsx
@@ -416,11 +416,11 @@ export class Plugins extends Component {
                             <p style={{ fontSize: '1.25em', margin: '0', color: 'black', fontWeight: '600' }}>
                                 {pluginsCount} plugins found
                             </p>
-                            Showing {paginationOffset + 1} to {' '}
+                            Showing {Math.min(pluginsCount, paginationOffset + 1)} to {' '}
                             {
                               // eslint-disable-next-line no-nested-ternary
-                              (paginationOffset + paginationLimit > plugins.totalCount) ?
-                                plugins.totalCount
+                              (paginationOffset + paginationLimit > pluginsCount) ?
+                                pluginsCount
                                 :
                                 (paginationOffset > 0) ?
                                   paginationOffset
```
Screenshot from 2022-04-09 15-37-54

somya51p:
@jennydaman Please review, I have tested for the plugins it is working and added these outside the return statement for code quality improvement.

For lint disabling comment `// eslint-disable-next-line no-nested-ternary`
we can implement that way, but I don't think it's as good a way.

```
// no-nested-ternary rule disallows nested ternary expressions
    const pluginsLast=(paginationOffset + paginationLimit > pluginsCount) ?
    pluginsCount : paginationLimit;
    if (paginationOffset + paginationLimit > pluginsCount)
          pluginsLast = pluginsCount;
    else if (paginationOffset > 0)
          pluginsLast = paginationOffset;
    else
          pluginsLast = paginationLimit;
    
```

src/components/Plugins/Plugins.jsx
```diff
@@ -416,17 +422,7 @@ export class Plugins extends Component {
...
-                              // eslint-disable-next-line no-nested-ternary
-                              (paginationOffset + paginationLimit > plugins.totalCount) ?
-                                plugins.totalCount
-                                :
-                                (paginationOffset > 0) ?
-                                  paginationOffset
-                                  :
-                                  paginationLimit
-                            }
+                            Showing {Math.min(pluginsCount, paginationOffset + 1)} to {pluginsLast}
```
jennydaman, line 425: The only situation where `pluginCount < paginationOffset + 1` would be where `pluginCount == 0`, in which case, it might be better to instead conditionally render something like "No plugins shown" or just omit this text instead. If we can get rid of this `Math.min` call, the what and why of this line of code would be more apparent by a quick glance.

src/components/Plugins/Plugins.jsx
```diff
@@ -369,7 +369,13 @@ export class Plugins extends Component {
     )
 
     const pluginsCount=plugins.totalCount > 0 ? plugins.totalCount : 0;
-    
+    const pluginsLast=(paginationOffset + paginationLimit > pluginsCount) ?
```
jennydaman, line 372: Nested ternary is hard to understand and there seems to be logic branch redundancy, I think this would do the same thing as:

```javascript
const pluginsLast = Math.min(paginationOffset + paginationLimit, pluginsCount);
```

- if no plugins, then `pluginsCount = 0`, so `pluginsLast = 0` :heavy_check_mark: 
- if on last page, then `paginationOffset + paginationLimit >= pluginsCount`, so `pluginsLast = pluginsCount` :heavy_check_mark: 
- if not on last page (implying there are multiple pages), then `paginationOffset + paginationLimit < pluginsCount`, so `pluginsLast = paginationOffset + paginationLimit` :heavy_check_mark:

jennydaman (changes requested):

src/components/Plugins/Plugins.jsx
somya51p: I got you! I will try to implement what you have suggested asap.

somya51p:
@jennydaman I have tried to implement the logic which you have suggested. Please have a look into it and suggest changes if any.
Screenshot from 2022-04-30 20-32-25
Screenshot from 2022-04-30 20-34-13

jennydaman:
We should fix this in the upstream client library https://github.com/FNNDSC/fnndsc/issues/76

PR #270, opened 2022-04-07
update readme file

sherrif10:
cc @mairin

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,10 @@ $ docker run --rm -it -v $(pwd):/home/localuser -p 3000:3000 -u $(id -u):$(id -g
 
 Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
 
+Login credentials: 
+username: .....
+password:....
+
 #### Precommit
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,10 @@ $ docker run --rm -it -v $(pwd):/home/localuser -p 3000:3000 -u $(id -u):$(id -g
 
 Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
 
+Login credentials: 
+username: .....
```
sherrif10, line 69: Hey @mairin , this is where we can add defualt password a user can simply use without creating a new one. 
If there is no default password, probably we can add that.

jennydaman: @sherrif10 it is not appropriate to specify a default username and password here. The ChRIS_store_ui does not know anything about what users exist in the backend (https://github.com/FNNDSC/ChRIS_Store), only the backend itself knows that.

The backend should not have default users, just like how there's no default user for Github.com. People who do not currently have an account can create one.

sherrif10: Make sense. thanks a lot much, the idea is its giving me hard time to login especially when i lost password.

sherrif10:
Close due to discussions made .

PR #271, opened 2022-04-07
Issue 244

somya51p:
@jennydaman Added yarn test with Github Actions to check if the` yarn` is built successfully, ignoring the test results.
Please check if this would work.


.github/workflows/yarn.yml
```diff
@@ -0,0 +1,24 @@
+name: yarn
```
jennydaman, line 1: It might be better to include this as a separate job of `ci.yml`, as a job called `test`, and then create a job dependency so that the container image build step depends on `test`. This would prevent non-working builds from being published.

jennydaman (changes requested):
Looks simple, and simple is good! It's great that you are looking into CI.

Please see comments.

.github/workflows/ci.yml
```diff
@@ -13,6 +13,16 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      
+      - name: test
```
jennydaman, line 17: the name "test" does not seem to be describing `actions/checkout@v2`

.github/workflows/ci.yml
```diff
@@ -13,6 +13,16 @@ jobs:
...
+        uses: actions/checkout@v2
+        uses: borales/actions-yarn@v2.3.0
```
jennydaman, line 19: From the README:

> Please keep in mind that this Action was originally written for GitHub Actions beta (when Docker was the only way of doing things). Consider using [actions/setup-node](https://github.com/actions/setup-node) to work with Yarn. This repository will be mostly supporting the existing flows.

https://github.com/Borales/actions-yarn

jennydaman (changes requested):

somya51p:
@jennydaman Can you please review the changes I recently made? I need a bit of help with this issue.

somya51p:
@jennydaman As discussed in the meeting, I reviewed the test results based on previous commits and concluded that the latest changes would work accurately. Please review the same and let me know if I am wrong.

PR #276, opened 2022-04-14
added copyrights to the signin page

AryaArunB:
Screenshot 2022-04-14 at 5 05 26 PM

### DIFF
--- a/src/components/SignIn/SignIn.jsx
+++ b/src/components/SignIn/SignIn.jsx
@@ -150,10 +150,17 @@ export class SignIn extends Component {
                   <Link to="/quickstart" href="/quickstart">
                     Signup
                   </Link>
+                   <div className="footer-copyright">
+      © 2018 - {new Date().getFullYear()} Boston Children&apos;s Hospital, Red
+      Hat, Massachusetts Open Cloud, Boston University. All rights reserved.
+    </div>
                 </p>
+                
               </Form>
+              
             </CardBody>
           </Card>
+         
         </div>
       </div>
     );


AryaArunB:
@jennydaman, @PintoGideon please review this.

src/components/SignIn/SignIn.jsx
```diff
@@ -150,10 +150,17 @@ export class SignIn extends Component {
                   <Link to="/quickstart" href="/quickstart">
                     Signup
                   </Link>
+                   <div className="footer-copyright" style={{color:"white"}}>
```
jennydaman, line 153: Using the CSS class `footer-copyright` here is a bad practice, while it's true that CSS classes are global, we would like to have some notion of scope. I would suggest either moving the class out of `src/components/Footer/Footer.css` and to some other CSS file which is explicitly designated to contain global classes, or copy the class to `SignIn.css`

jennydaman: Scoping would protect against a fix for https://github.com/FNNDSC/ChRIS_store_ui/issues/151 (which enforces scoping)

jennydaman (changes requested):

src/components/SignIn/SignIn.jsx
AryaArunB: Thanks for the feedback @jennydaman I have copied the class to SignIn.css

AryaArunB:
@jennydaman is there something more that you'd like me to do in this pr?

Issue #167, opened 2021-04-25
Use CSS-in-JS from @patternfly/react-styles


jennydaman:
https://www.npmjs.com/package/@patternfly/react-styles

Continues https://github.com/FNNDSC/ChRIS_store_ui/pull/162

somya51p:
@jennydaman Can I look into this one too. Can you help me out finding the resources for [@pattern/react-styles](https://www.npmjs.com/package/@patternfly/react-styles)

DivyanshiSingh:
@jennydaman I would like to take this up next.

jennydaman:
@somya51p as mentioned in our conversation in #162, I am not sure where to find documentation on @patternfly/react-styles either. It'll probably take a skim of the source code and a good IDE.

@DivyanshiSingh this issue depends on #162 which has not been merged yet.

Akpjunior94:
Hi @mairin, My Name is Akpan Abraham, I'm an Outreachy intern and want to contribute to the project "Create external app registration system for React-based user interface". Can I be assigned to this issue please?

jennydaman:
Hey everyone sorry for missing these replies!

The issue is about using the "CSS-in-JS" syntax so that CSS files are imported as JavaScript objects, instead of referring to CSS classes by name using strings. For example, instead of


```jsc
import 'styles.css';
...
<div className="something" />
```


We prefer

```jsx
import { styles } from 'styles.css';
...
<div className={styles.something} />
```

The advantage of CSS modules are, class names can be scoped and they can also be checked to exist at build time.

mairin:
This issue is still open, see https://github.com/FNNDSC/ChRIS_store_ui/issues/167#issuecomment-957889957 for details on how to complete this.

somya51p:
> This issue is still open, see [#167 (comment)](https://github.com/FNNDSC/ChRIS_store_ui/issues/167#issuecomment-957889957) for details on how to complete this.

@mairin I would like to work on this issue. Please assign me the same.
ThankYou!!

mairin:
@somya51p Sure I just assigned to you.

Mfonn:
@jennydaman I will like to work on this issue.

somya51p:
Hey @Mfonn, I am already working on it :)

Mfonn:
Okay then, I'll find another issue to work on.

somya51p:
[...]

PR #277, opened 2022-04-16
Added css to css module script

somya51p:
@jennydaman I have created the script to convert CSS to CSS modules using regular expressions in js. Please review the PR and suggest changes if any.


sherrif10:
@somya51p  What exactly does this do. How can i test this generally.

somya51p:
> @somya51p What exactly does this do. How can i test this generally.

Hey, @sherrif10  It's associated with issue #167. This script would automatically convert CSS files to CSS modules.

scripts/cssToCssModule.js
```diff
@@ -0,0 +1,79 @@
+const fs = require("fs");
+const path = require("path");
+
+const blackListedClasses = [
```
jennydaman, line 4: these are from patternfly right?

scripts/cssToCssModule.js
```diff
@@ -0,0 +1,79 @@
...
+const arrayOfFiles = filesToChange(pathToComponents);
+
+function convertFile(filePath) {
+  fs.readFile(filePath, "utf8", (err, textData) => {
+    const cssFileName = filePath.replace(".jsx", ".css");
+    const cssModuleName = filePath.replace(".jsx", ".module.css");
+    fs.rename(cssFileName, cssModuleName, () => {});
+    let data = textData.replace(
+      `import './${cssFileName.split("/").slice(-1)}'`,
+      `import styles from './${cssModuleName.split("/").slice(-1)}'`
```
jennydaman, line 50: What happens if multiple stylesheets are imported in the same JSX file? (Or how can you verify that this is not a concern?)

scripts/cssToCssModule.js
```diff
@@ -0,0 +1,79 @@
...
+      if (className.length > 1) {
+        res = "{`";
+        className.forEach((cls, index) => {
+            if (Number(index) !== 0) {
+                res += " ";
+              }
+              if (blackListedClasses.includes(cls)) {
+                  res += cls;
+              } else {
+                  res += `\${styles[${  cls  }]}`;
```
jennydaman, line 64: do we need quotes around cls?

jennydaman (commented):
Disclaimer: I haven't tried running this yet, I will try later.

scripts/cssToCssModule.js
somya51p: > these are from patternfly right?

Yes, both bootstrap and patternfly

scripts/cssToCssModule.js
somya51p: > What happens if multiple stylesheets are imported in the same JSX file? (Or how can you verify that this is not a concern?)

I need to check that for such cases. As in the current codebase, such a thing does not persist so it would work fine ig. The script would convert the files to CSS modules successfully but I cant say about the impact at this very moment.

scripts/cssToCssModule.js
somya51p: > do we need quotes around cls?

don't need that for normal `className`

scripts/cssToCssModule.js
jennydaman: Okay.

It would be helpful if you could provide a way to _verify_ this is not a concern, i.e., explain how you did your "analysis."

I looked into this and here's a script I ran:

```shell
fd -e jsx -x bash -c '[ $(grep -F .css {} | grep -v node_modules | wc -l) -lt 2 ] || echo "WARNING: {} imports more than 1 local CSS file."' src
```

This script translates to: "for all JSX files under the `src/` directory, count the number of lines which include the substring ".css" but do not include the substring "node_modules", and if this number is greater than or equal to 2, print a warning."

scripts/cssToCssModule.js
jennydaman: During our meeting we talked about how we do need quotes here. I suggest you write examples and a unit test for this case.

scripts/cssToCssModule.js
somya51p: > Okay.
> 
> It would be helpful if you could provide a way to _verify_ this is not a concern, i.e., explain how you did your "analysis."
> 
> I looked into this and here's a script I ran:
> 
> ```shell
> fd -e jsx -x bash -c '[ $(grep -F .css {} | grep -v node_modules | wc -l) -lt 2 ] || echo "WARNING: {} imports more than 1 local CSS file."' src
> ```
> 
> This script translates to: "for all JSX files under the `src/` directory, count the number of lines which include the substring ".css" but do not include the substring "node_modules", and if this number is greater than or equal to 2, print a warning."

Ok, will take care of that, and also write unit tests. Thanks for the review.

somya51p:
@jennydaman Please look into the changes made and tests created.

somya51p:
Hey @jennydaman, to be precise here are the changes that I had made in the latest commits of this PR. 
1. Created helper functions and added comments.
2. Added quotes for single classNames. 
3. Removed bug for multiple blacklisted classes.too
4. Wrote tests for all the different test cases possible.

Please review it and let me know if I am missing out something. Thankyou!

src/scripts/cssToCssModule.js
```diff
@@ -0,0 +1,162 @@
...
+
+    files.forEach((file) => {
+        if (fs.statSync(`${dirPath  }/${  file}`).isDirectory()) {
+            arrayOfFiles = filesToChange(`${dirPath  }/${  file}`, arrayOfFiles);
+        } else if (file.split(".").slice(-1)[0] === "jsx") {
+            arrayOfFiles.push(path.join(dirPath, "/", file));
+        }
+    });
+    return arrayOfFiles;
+};
```
jennydaman, line 46: Here's how I would've done it:

```javascript
const fs = require('fs');
const path = require('path');

/**
 * @param {string} p an existing path
 * @returns list of all subpaths under p, including itself
 */
function findAllPaths(startingPath) {
  if (fs.statSync(startingPath).isDirectory()) {
    const subpaths = fs.readdirSync(startingPath).map((subpath) => {
      return findAllPaths(path.join(startingPath, subpath));
    });
    return Array.prototype.concat.apply([startingPath], subpaths);
  }
  return [startingPath];
}

// this helper function here has BOTH a self-explanatory name AND a simple implementation,
// so I am choosing to not add documentation
function filterByFileExtension(paths, suffix) {
  return paths.filter((p) => {
    return fs.statSync(p).isFile() && p.endsWith(suffix);
  });
}

// example

const exampleDirectory = 'ChRIS_store_ui/src';

filterByFileExtension(findAllPaths(exampleDirectory), '.jsx').forEach(p => {
  console.log(p);
});
```

I've identified the individual units of functionality of this task and created a helper for each one. This way, each function is individually testable, so if either one is doing the wrong thing, it's much easier to identify where the bug is in my code, and also the fix for my code will be much smaller. It can also be easier for others to read, understand, and review my code, because the names of my functions are descriptive and there is docstring where necessary.

jennydaman (changes requested):
Hi @somya51p 

I am still overwhelmed with a backlog of work but I see that you have worked very hard on this so I am providing some initial feedback now on this weekend. I will add more comments later.

somya51p:
> Hi @somya51p
> 
> I am still overwhelmed with a backlog of work but I see that you have worked very hard on this so I am providing some initial feedback now on this weekend. I will add more comments later.

Ohk, Thanks a lot for taking out time and reviewing it :)
I will study your comments and implement them in the next commit.

PR #280, opened 2022-04-17
Refactored docker run command

### LINKED ISSUE #278: Docker run throws error
On starting the ChRIS store UI development server using docker on Ubuntu 20.04, `docker run` throws error
`invalid reference format: repository name must be lowercase
`

archijaiswal:
Fixes #278
Added quotation to the path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Open a new terminal and type:
 ```bash
 $ git clone https://github.com/FNNDSC/ChRIS_store_ui.git
 $ cd ChRIS_store_ui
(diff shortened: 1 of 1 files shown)


jennydaman (approved):

jennydaman:
@archijaiswal just some notes:

This does not fix #278. The error `docker` gave you was correct but perhaps misleading. You probably have a space in the path of your current working directory. Generally it's discouraged to have spaces in paths, but this change you've made is nonetheless an improvement.

This change is not a "refactor." A refactor is a change to the code-base which does not affect behavior. Your change does affect behavior (in a desirable way).

archijaiswal:
Oh, alright! Noted. Thanks a lot @jennydaman for such a detailed explanation and your guidance, I really appreciate it. I'll be much aware of these things from now on.

Issue #206, opened 2021-05-26
Prevent application crashes with error boundaries


zrthxn:
https://reactjs.org/docs/error-boundaries.html

TateDeVito:
Hi @zrthxn - I could try to take a look at this one. Could you give any more detail on specific components you would like to apply the Error Boundary to? Any class of errors you are most trying to prevent?

zrthxn:
@TateDeVito 

The application was susceptible to crashes usually caused by "cannot read property ... of undefined/null" errors. I took care to make sure, as best I could, to prevent these in #229 and #231. However this is not completely guaranteed as this is a JS app and we can't properly fix these things ahead-of-time. 

The error boundary needs to mostly prevent the following types of errors
- errors caused by accessing properties on component "props" which are `undefined` or `null`,
- errors caused by unexpected API responses, like `null` instead of data,
- errors caused by infinite loops or infinite re-renders.

Currently what we have is a `showNotification` function in every component that might have an error and we `catch` where we think an error might arise, and then call that function. What I wanted to do was make a global, application wide error system wherein any component that errored-out would register that error message and then a notification can be shown or some action taken (maybe log or report if its caused by a logical error?).
**The goal is that the user should never have to look at a blank errored-out application.**

The following components are the ones where errors are likely to happen (not that they currently do)
- [Plugin.jsx](https://github.com/FNNDSC/ChRIS_store_ui/blob/master/src/components/Plugin/Plugin.jsx)
- [PluginMeta.jsx](https://github.com/FNNDSC/ChRIS_store_ui/blob/master/src/components/Plugin/PluginMeta.jsx)
- [Plugins.jsx](https://github.com/FNNDSC/ChRIS_store_ui/blob/master/src/components/Plugins/Plugins.jsx)
- [Dashboard.jsx](https://github.com/FNNDSC/ChRIS_store_ui/blob/master/src/components/Dashboard/Dashboard.jsx)
- [CreatePlugin.jsx](https://github.com/FNNDSC/ChRIS_store_ui/blob/master/src/components/CreatePlugin/CreatePlugin.jsx) 

Further details and advice on the direction of this whole application can be taken from @jennydaman. And you can also talk to me about any of the code I've written.

TateDeVito:
Ok thanks @zrthxn . I will take a look at this! Will get back if I have any questions/other items.

TateDeVito:
[...]

PR #285, opened 2022-04-20
feat(app): implement a global error boundary component

roy98:
**This Pull Request is associated with this issue #206** 

- Implement an error boundary component to handle errors that can corrupt the internal state of React.

- Display a fallback UI to the user

image


roy98:
@jennydaman can you please have a look?

src/components/ErrorBoundary/ErrorBoundary.jsx
```diff
@@ -0,0 +1,69 @@
...
+              <h1>400</h1>
+              <h2>{this.state.error?.message}</h2>
+              <p>We are having an issue, please click on the button bellow to reload the page !</p>
+
+              <Button id="reload-error-btn" variant="primary" onClick={() => this.reload()}>
+                <span>Retry</span>
+              </Button>
+            </div>
+            <div className="illustration-container">
+              <img src={Illustration} alt="Error illustration" className="illustration" />
```
jennydaman, line 58: Where did this illustration come from? Just want to make sure reuse is permitted.

Or perhaps (in a future issue) we should loop back with @mairin and consider using ["brainy" the mascot](https://github.com/FNNDSC/cube-design/blob/master/ChRIS-Store/mascot_expressions.svg) here (similar to the 404 error page).

src/components/ErrorBoundary/ErrorBoundary.jsx
```diff
@@ -0,0 +1,69 @@
...
+            closeable
+            onClose={() => {}}
+          />
+          <div className="container">
+            <div className="error-message-container">
+              <h1>400</h1>
+              <h2>{this.state.error?.message}</h2>
+              <p>We are having an issue, please click on the button bellow to reload the page !</p>
+
+              <Button id="reload-error-btn" variant="primary" onClick={() => this.reload()}>
```
jennydaman, line 53: Currently, the `reload` function doesn't necessarily reload the page but navigates us back to `/`. Do you think it would be better to try reloading the page, or navigating "home"? In the latter case, we should instead say here "please click on the button below to go home" (or maybe go back)?

src/components/ErrorBoundary/ErrorBoundary.jsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  showNotifications = (error) => {
+    this.setState((prev) => ({
+      errors: [...prev.errors, error],
+    }));
+  };
+
+  reload = () => {
+    window.location.href = '/';
```
jennydaman, line 31: Pedantic: don't assume web app is served from `/`, i.e. this will not work if deployed on a subpath [(by setting `PUBLIC_URL` or `homepage`)](https://create-react-app.dev/docs/advanced-configuration/).

src/components/ErrorBoundary/ErrorBoundary.jsx
```diff
@@ -0,0 +1,69 @@
...
+            title="Error"
+            message={this.state.error?.message || 'Something happened!'}
+            position="top-right"
+            variant="danger"
+            closeable
+            onClose={() => {}}
+          />
+          <div className="container">
+            <div className="error-message-container">
+              <h1>400</h1>
```
jennydaman, line 49: Not necessarily a 400 error, could be a client-side error.

src/components/ErrorBoundary/ErrorBoundary.jsx
```diff
@@ -0,0 +1,69 @@
...
+              <h2>{this.state.error?.message}</h2>
+              <p>We are having an issue, please click on the button bellow to reload the page !</p>
```
jennydaman, line 51: Please proofread.

Idea: we should provide a link to https://github.com/FNNDSC/ChRIS_store_ui/issues

src/components/ErrorBoundary/ErrorBoundary.jsx
```diff
@@ -0,0 +1,69 @@
...
+  reload = () => {
+    window.location.href = '/';
+  };
+
+  render() {
+    if (this.state.hasError) {
+      return (
+        <>
+          <ErrorNotification
+            key={`notif-boundary`}
```
jennydaman, line 39: Does our eslint config allow for unnecessary string template? If so, I should change that.

src/components/ErrorBoundary/ErrorBoundary.jsx
```diff
@@ -0,0 +1,69 @@
...
+            title="Error"
+            message={this.state.error?.message || 'Something happened!'}
+            position="top-right"
+            variant="danger"
+            closeable
```
jennydaman, line 44: Is `Notification` component to use here? If an uncaught exception happens in a child component, I don't think a closable notification would be suitable, we would probably want the entire view to be replaced by an error screen (stop the user from interacting with the program when its state becomes undefined/erroneous)

jennydaman (changes requested):

src/components/ErrorBoundary/ErrorBoundary.jsx
roy98: The illustration come from [undraw](https://undraw.co/)

src/components/ErrorBoundary/ErrorBoundary.jsx
roy98: > Currently, the `reload` function doesn't necessarily reload the page but navigates us back to `/`. Do you think it would be better to try reloading the page, or navigating "home"? In the latter case, we should instead say here "please click on the button below to go home" (or maybe go back)?

I would be better to navigate "home". I got your comments

src/components/ErrorBoundary/ErrorBoundary.jsx
roy98: Ok, I will do it then

PR #297, opened 2022-10-14
Fix: Single quote rule is not enforced 

Seundede:
Enforced the use of single quotes
After:
Screenshot 2022-10-14 at 16 20 29

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,9 +1,5 @@
 {
-  "extends": [
-    "plugin:react/recommended",
-    "airbnb",
-    "prettier"
-  ],
+  "extends": ["plugin:react/recommended", "airbnb", "prettier"],
   "env": {
     "node": true,
     "jest": true,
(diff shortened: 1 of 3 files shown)


jennydaman:
I was able to address the issue with a smaller patch:

```patch
diff --git a/.eslintrc b/.eslintrc
index 7329dbf..4a7b77d 100644
--- a/.eslintrc
+++ b/.eslintrc
@@ -32,6 +32,7 @@
     "prettier"
   ],
   "rules": {
+    "quotes": ["error", "single"],
     "react/jsx-filename-extension": [1, { "extensions": [".jsx"] }],
     "import/no-extraneous-dependencies": ["error", {"devDependencies": ["**/*.test.js", "**/*.spec.js", "./src/setupTests.js"]}],
     "react/state-in-constructor": "off",
@@ -39,4 +40,4 @@
     "react/require-default-props": "off",
     "react/jsx-props-no-spreading": "off"
   }
-}
\ No newline at end of file
+}
```

This might be a preferable solution, but I was hesitant about adding this one line of code myself because I don't understand why the rule is not inherited from the airbnb and prettier rulesets. Do you have an idea for why this may be?

Seundede:
I think it's because `eslint-config-prettier` disables all rules that might conflict with prettier.

jennydaman (approved):

PR #309, opened 2022-10-28
Bugfix plugins counter / paginationOffset

thessakranendonk:
Issue #268 

I know this issue got worked on 6 months ago but it's not merged yet and the issue hasn't been worked on since.

Before:
Screen Shot 2022-10-28 at 1 32 26 PM

- Resolved the issue of not showing paginationOffset correctly
- Resolved showing 1 to -1 or 1 tot 1 when there aren't any plugins


jennydaman:
We should fix in upstream client library. https://github.com/FNNDSC/fnndsc/issues/76

PR #310, opened 2022-10-29
Fix: Plugin dropdown menu gets cut off on mobile

### LINKED ISSUE #294: Plugin drop down menu gets cutoff on mobile screen
Screenshot from 2022-10-12 20-42-54

I noticed this issue and fixed it.

Screenshot from 2022-10-12 21-13-02

Chizycodes:
Closes #294

Before: 
dropdown-fix-before

After: 
dropdown-fix-after

### DIFF
--- a/src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.jsx
+++ b/src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.jsx
@@ -150,15 +150,30 @@ class DashPluginCardView extends Component {
                 <Dropdown
                   className="card-view-kebob"
                   onSelect={(event) => this.onSelect(event, plugin)}
-                  toggle={<KebabToggle onToggle={(value) => this.toggleEditMenu(value, index)} id={`kebab-${plugin.id}`}/>}
+                  toggle={
+                    <KebabToggle
+                      onToggle={(value) => this.toggleEditMenu(value, index)}
+                      id={`kebab-${plugin.id}`}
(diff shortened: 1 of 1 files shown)


src/components/Dashboard/components/DashPluginCardView/DashPluginCardView.jsx
```diff
@@ -150,15 +150,30 @@ class DashPluginCardView extends Component {
...
                   onSelect={(event) => this.onSelect(event, plugin)}
-                  toggle={<KebabToggle onToggle={(value) => this.toggleEditMenu(value, index)} id={`kebab-${plugin.id}`}/>}
+                  toggle={
+                    <KebabToggle
+                      onToggle={(value) => this.toggleEditMenu(value, index)}
+                      id={`kebab-${plugin.id}`}
+                    />
+                  }
                   isOpen={isOpen[index]}
+                  position={window.innerWidth <= 720 ? 'right' : 'left'}
```
jennydaman, line 160: I wonder if a CSS mediaquery could be a better solution? The advantage of your approach in the JSX is that it leverages the `position` prop of `<Dropdown>` though it's not responsive (try opening the dropdown then resizing the screen width).

This is just a nitpick though, I like your solution,

jennydaman (approved):
Ideally, line 160 is the only necessary change. It's a good practice to not refactor (formatting) code unrelated to the feature. Nonetheless, good work!

PR #314, opened 2022-11-13
Fix close button of the notification does not work and modernize notification pop up

### LINKED ISSUE #284: The close button of the notification does not work
### The close button of the notification does not work
Screen Shot 2022-04-20 at 23 53 46

thessakranendonk:
PR for issue #284 : The close button of the notification does not work

- Fixed not closing of alert notification button.
- Added a new Popup.jsx component that looks more modern and is more prominent on the page. This popup looks more like a modal and makes the background of the page slightly darker when the notification pops up. Please let me know if this component would be a good replacement for the current notification or if there are any UI improvement suggestions..
- In Issue 284 there was some discussion of removing the ```closeable``` prop but I think this prop is useful here. If the prop is passed to the component it will automatically close a notification after a certain timeframe without the user having to click the close button. Personally I don't like having to close popups by mouse click only and therefore I think this prop should stay.

Before:
Screen Shot 2022-11-13 at 6 18 48 PM

After:
Screen Shot 2022-11-13 at 6 20 30 PM
Screen Shot 2022-11-13 at 6 20 46 PM
Screen Shot 2022-11-13 at 6 21 27 PM

### DIFF
--- a/src/components/Notification/Notifications.css
+++ b/src/components/Notification/Notifications.css
@@ -30,7 +30,7 @@
 }
 
 .top-left {
-    top: 12px;
+    top: 100px;
     left: 12px;
     transition: transform .3s ease-in;
     animation: notification-in-down .3s;
--- a/src/components/Notification/Popup.css
+++ b/src/components/Notification/Popup.css
@@ -1,0 +1,160 @@
+.modal_wrapper {
+  z-index: 2999;
+  width: 100vw;
+  height: 100vh;
+  background-color: rgba(0, 0, 0, 0.338);
+  position: absolute;
+  left: 0;
+  top: 0;
+}
+
+.modal_container {
+  position: fixed;
+  z-index: 3000;
+  transition: .3s ease;
+  display: flex;
+  justify-content: space-between;
+  height: max-content;
+  border-radius:12px;
+  box-shadow: 0 0 10px rgb(85, 84, 84);
+}
+
+.modal {
+  display: flex;
+  flex-direction: column;
+  width: 35vw;
+  height: auto;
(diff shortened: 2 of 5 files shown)


jennydaman:
Hey @thessakranendonk this is impressive. I will review the design first and then peek at your code in a later comment.

Before implementing changes to design or other major features, I recommend doing mock-ups first so that we can discuss before investing effort.

The most striking design element to discuss here is color choice, but also we can talk a bit about border radius and iconography. The ChRIS project uses the _Patternfly v4_  design language and you should refer to its specification here:

https://www.patternfly.org/v4/guidelines/colors

- The colors are attention seeking, but I don't think they fit Patternfly's color palette.
- I don't think these notifications are important enough to warrant dimming the screen plus a popup.

You've done a lot to make the popups as noticeable as can be, but it comes at the tradeoff of a cohesive design language.

IMO header bars should not be used on pop-ups. Header bars are implemented for window managers of desktop operating systems for reasons such as providing a handle for window placement, displaying windows buttons such as close, and the window title. Here, there is only one functionality (providing space for the close button) so the header bar is wastefully taking up a lot of space (with a striking color).

For more information about header bars, my favorite design spec for them is the GNOME HIG. https://developer.gnome.org/hig/patterns/containers/header-bars.html

jennydaman:
Tagging @mairin, you are always welcome to chat with us about design in [#chris-frontend:fedora.im](https://matrix.to/#/#chris-frontend:fedora.im)

thessakranendonk:
@jennydaman thank you for your feedback!! I will have another look at this and make some changes. Very helpful! Thank you!

